PR #64, opened 2021-09-15
changes condition to effect in y-axis labels

resulumit:
The default y-axis labels in average/conditional marginal effect plots reads as "Marginal effect of _condition_variable_". I think this is meant to be "Marginal effect of _effect_variable_" instead.


vincentarelbundock:
Good catch! Thanks so much for taking the time to fix this!

FYI, I'm still working a lot on this, and am recruiting some help to cross-validate all the numerical results. If everything goes according to plan, there should be a first CRAN release in a few weeks.

Issue #226, opened 2022-03-16
Three-way interaction using factor variables: Results differ from margins


jolyphil:
Thanks for this excellent and very useful package (again)!

It seems like margins and marginaleffects produce different results when calculating marginal effects based on a model that includes a three-way interaction with factor variables. 

Note that, in the example below, the divergence appears _only_ in a **three-way interaction** when both `vs` and `am` are defined as **factor variables**. 

The solution offered by margins is closer to the results from Stata. 

Any idea what is happening here? 

Thanks a lot!

``` r
# Set up ------------------------------------------------------------------

library(margins)
library(marginaleffects)
library(dplyr)

df <- mtcars

df$vs <- as.factor(df$vs)
df$am <- as.factor(df$am)

hp_values <- c(100, 200, 300)


# Two-interaction: Same results -------------------------------------------

m1 <- lm(mpg ~ am * hp + vs, data = df)

margins(
  m1,
  variables = "am",
  at = list(hp = hp_values)) |>
  summary() |>
  select(AME)
#>    AME
#>  5.858
#>  4.688
#>  3.518

marginaleffects(
  m1,
  variables = "am",
  newdata = datagrid(hp = hp_values)) |>
  select(dydx)
#>       dydx
#> 1 5.857547
#> 2 4.687794
#> 3 3.518041


# Three-way interaction: different results --------------------------------

m2 <- lm(mpg ~ am * hp * vs, data = df)

margins(
  m2,
  variables = "am",
  at = list(hp = hp_values)) |>
  summary() |>
  select(AME)
#>    AME
#>  4.386
#>  4.171
#>  3.955

marginaleffects(
  m2,
  variables = "am",
  newdata = datagrid(hp = hp_values)) |>
  select(dydx)
#>       dydx
#> 1 3.709359
#> 2 4.239618
#> 3 4.769877
```

vincentarelbundock:
On my phone now so I can't run code to check, but I am 96% sure that the difference is the same thing that I described here:

https://github.com/strengejacke/ggeffects/issues/249#issuecomment-1068334112

In that example, I talk about the `prediction` package, which was written by the `margins` author and behaves similarly.

jolyphil:
Thanks a lot for the clarification! That's really useful. 

I now get the same estimates with the argument `grid.type = "counterfactual"` but the standard errors are still different. 

``` r
library(margins)
library(marginaleffects)
library(dplyr)

df <- mtcars

[...]

PR #231, opened 2022-03-29
Add Stata vignette

jolyphil:
Here is a first attempt at creating a vignette comparing Stata's margins command and the functions of marginaleffects. I was inspired by these two issues:

https://github.com/vincentarelbundock/marginaleffects/issues/226
https://github.com/strengejacke/ggeffects/issues/249

And the margins' documentation here:

https://thomasleeper.com/margins/articles/Stata.html

### DIFF
--- a/vignettes/stata.Rmd
+++ b/vignettes/stata.Rmd
@@ -1,0 +1,190 @@
+---
+title: "Comparison with Stata's `margins` command" 
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Comparison with Stata's `margins` command}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  fig.width = 6,
+  fig.asp = .4,
+  warning = FALSE,
+  message = FALSE,
+  comment = "#>"
+)
+```
+
+## Average Marginal Effect (AMEs)
+
+### Stata
+
+Both Stata's `margins` command and the `marginaleffects` function can calculate average marginal effects (AMEs). Here is an example showing how to estimate AMEs in Stata: 
+
+```
+quietly reg mpg cyl hp wt
+margins, dydx(*)
+
+Average marginal effects                        Number of obs     =         32
+Model VCE    : OLS
+ 
+Expression   : Linear prediction, predict()
+dy/dx w.r.t. : cyl hp wt
+ 
+------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


vignettes/stata.Rmd
```diff
@@ -0,0 +1,190 @@
...
+In other words, average counterfactual adjusted predictions as implemented by Stata are a hybrid between predictions at the observed values (the default in `marginaleffects::predictions`) and predictions at representative values. 
+
+### marginaleffects
+
+You can estimate average counterfactual adjusted predictions with `predictions()` by, first, setting the `grid.type` argument of `data.grid()` to `"counterfactual"` and, second, by averaging the predictions, for example using `aggregate()` or `dplyr::summarise()`.
+
+```{r}
+mod <- lm(mpg ~ as.factor(cyl) * hp + wt, data = mtcars)
+
+nd <- datagrid(model = mod,
```
vincentarelbundock, line 180: If you use `datagrid()` inside the function call directly with the `newdata` argument, you won't need to specify the `model` argument manually. Also, if you specify inside function call, you probably don't need the `as.factor()`, since `marginaleffects` will coerce it to factor if it sees that it is a factor in the original data.

vignettes/stata.Rmd
```diff
@@ -0,0 +1,190 @@
+---
+title: "Comparison with Stata's `margins` command"
```
vincentarelbundock, line 2: This will eventually be a "Alternative software packages" page, I think, not just Stata.

vignettes/stata.Rmd
```diff
@@ -0,0 +1,190 @@
...
+  3. It takes the average marginal effect for each value of cyl. 
+
+### marginaleffects
+
+You can estimate average counterfactual marginal effects with `marginaleffects()` by, first, setting the `grid.type` argument of `data.grid()` to `"counterfactual"` and, second, by averaging these effects, for example using `aggregate()` or `dplyr::summarise()`. 
+
+```{r}
+mod <- lm(mpg ~ as.factor(cyl) * hp + wt, data = mtcars)
+
+nd <- datagrid(model = mod,
```
vincentarelbundock, line 108: See comment below. I think it's more convenient to use `datagrid` in the main function call with the `newdata` argument.

vincentarelbundock:
This is awesome, thanks so much!

I left a couple comments, but all really minor as I think the vignette hits the key points.

Let me know when you think it's ready to merge and I'll hit the button.

Thanks again!

vincentarelbundock:
Merged and renamed to "alternative_software.Rmd"

Thanks again, this is great!

jolyphil:
Great! Thanks a lot!

PR #400, opened 2022-07-07
Stata vs marginaleffects elasticities

marco-mendoza-avina:
Stata margins using dydx, eydx, dyex with `lm` and `glm` models.


vincentarelbundock:
Merci, c'est parfait!

PR #454, opened 2022-08-14
Unit-level contrasts vignette v1

arthur-albuquerque:


vincentarelbundock:
Thanks @arthur-albuquerque, this looks great\!

@ngreifer, should we have a `oravg` shortcut?

``` r
library(marginaleffects)
library(rms)

load(url(
"https://github.com/vincentarelbundock/modelarchive/raw/main/data-raw/gusto.rda"
))

gusto <- subset(gusto, tx %in% c("tPA", "SK"))
gusto$tx <- factor(gusto$tx, levels = c("tPA", "SK"))

mod <- glm(
    day30 ~ tx + rcs(age, 4) + Killip + pmin(sysbp, 120) + lsp(pulse, 50) +
    pmi + miloc + sex, family = "binomial",
    data = gusto)

oravg <- \(hi, lo) (mean(hi) / (1 - mean(hi))) / (mean(lo) / (1 - mean(lo)))
comparisons(
    mod,
    variables = "tx",
    transform_pre = oravg) |>
    summary()
#>   Term Contrast Effect Std. Error z value   Pr(>|z|) 2.5 % 97.5 %
#> 1   tx   custom  1.192    0.05349   22.28 < 2.22e-16 1.087  1.296
#> 
#> Model type:  glm 
#> Prediction type:  response 
#> Pre-transformation:  oravg
```

vincentarelbundock:
Answering my own question: No, Vincent, just exponentiate the coefficient.

arthur-albuquerque:
Great edits, Vincent.

I saw that you removed the terms "adjusted" in the vignette. I had put those following section "2.3 Unadjusted and adjusted analyses" in [this article](https://onlinelibrary.wiley.com/doi/10.1002/bimj.201900297). I believe explicitly stating those marginal estimands are adjusted would be helpful. What do you thing?

arthur-albuquerque:
> Answering my own question: No, Vincent, just exponentiate the coefficient.

I would have actually voted for "yes". An `orgavg` would be analog to the current `ratioavg`, as it already is for `lnratioavg` and `lnoravg`.

vincentarelbundock:
I'm just not convinced that quantity makes much sense. In the same spirit, I think we may want to remove the `lnoravg` example from the vignette (but keep the adjusted risk ratio example).

I'm fine with adding back the word "adjusted". Feel free to make and push new commit.

arthur-albuquerque:
I would keep the odds ratio example because Dr. Harrell references to OR multiple times in this vignette.

vincentarelbundock:
Hmmm, I'm still not convinced. In the blog, OR is mentioned as a one-number summary, and I'm pretty sure he's referencing the usual OR. Since ORs are identical for each individual, I'm not sure how we should interpret an "OR computed using average predictions". That just seems like a weird quantity to me.

If someone can show me a textbook or reference in the scientific literature that makes a compelling case for that quantity, I'll be happy to add it back to the vignette and create a shortcut for it, but for now I'll keep this out.

Thanks again for your work on this, I really appreciate it!

arthur-albuquerque:
Thank you!

And sorry, I misunderstood what you meant; no need to discuss ORs in this vignette. 

FYI, Dr. Harrel has another [blog post](https://www.fharrell.com/post/varyor/) on logistic regressions with interactions where he does show patient-specific odds ratios (and risk ratios). He doesn't cite any peer-reviewed article, though.

image

In addition, there is this bayesian version of the same post [here](https://benyandrew.netlify.app/blog/bayesian_rct/#) by @andrew10043

vincentarelbundock:
Oh, that seems like an interesting blog post! Would be cool to eventually replicate those analyses as well, maybe using `plot_cco()`. But that's for another day :)

Thanks again.

ngreifer:
@vincentarelbundock I say definitely keep the marginal log OR. This is a commonly estimated quantity in randomized trials and other analyses with binary outcomes. See [here](https://onlinelibrary.wiley.com/doi/abs/10.1002/sim.2781) and [here](https://doi.org/10.1097/EDE.0000000000001370) for examples in the methodological literature where its properties are examined. I personally would never report it, but I still think it is an important option for users.

`"oravg"` I think you can do without because inference based on the delta method will be invalid except in massive samples. Normal theory inference should apply to `"lnoravg"` and the user can have the option of exponentiating the coefficients and confidence intervals with, e.g., a `transform_post`-like option (as of now, `transform_avg` in `summary()`). I feel the same about the risk ratio; the quantity is obviously not normally distributed, but its log more plausibly is.

arthur-albuquerque:
Hi @vincentarelbundock,

I know the default in `comparisons()` is `vcov = T`, thus it estimates 95% CIs for all unit-level contrasts through the delta method.

How can one interpret these unit-level 95% CIs? Dr. Harrell did not discuss these in his blog post.

Issue #485, opened 2022-09-12
Marginal effects on fixest glm model is slow


maellecoursonnais:
Hi, 

First, thank you for all the efforts you put into this very important package. I am trying to compute marginal effects and predicted probabilities from a pretty large fixed-effect logistic regression. The command, however, is very slow and is thus not really usable. Do you see a reason for that?

Minimal example:

```r
library(fixest)
library(marginaleffects)
df <- data.frame(a = rnorm(1e6),
                 b = runif(1e6),
                 c = sample(290, 1e6, replace = T), 
                 d = sample(c(0, 1), 1e6, replace = T),
                 e = sample(1:5, 1e6, replace = T))

fe <- feglm(d ~ a + b  + i(e) + i(e, b) | c, data = df, family = "binomial")

predictions(fe, newdata = datagrid(model = fe, a = quantile(df$a), b = quantile(df$b)))
```
---
Edit: 
The function actually stopped and threw this error:

> Error: 
> Unable to compute adjusted predictions for model of class `fixest`. You can try to
> specify a different value for the `newdata` argument. If this does not work and
> you believe that this model class should be supported by `marginaleffects`,
> please file a feature request on the Github issue tracker:
> 
> https://github.com/vincentarelbundock/marginaleffects/issues
> In addition: Warning message:
> 
> Unable to extract predictions from a model of type `fixest`. Please
> report this problem, along with replicable code, on the `marginaleffects` issue tracker:
> https://github.com/vincentarelbundock/marginaleffects/issues
>

vincentarelbundock:
This minimal example runs in a fraction of a second on my computer.

Could you try updating the following package to the devel versions, restarting R, and trying again?

```r
library(remotes)
install_github("lrberge/fixest")
install_github("easystats/insight")
install_github("vincentarelbundock/marginaleffects")
```

maellecoursonnais:
You are right, it indeed was fast. I am trying to reproduce the problem. I added categorical variables in the original model.

vincentarelbundock:
I believe that this is a problem in the `is_fixest_call` and `is_calling_fun` functions from the `fixest` package. These functions take a *long* time to compute when `fixest` methods are called programmatically instead of directly. Here’s a benchmark showing that it takes much longer when using `do.call` rather than `predict()` directly:

``` r
library(fixest)
library(microbenchmark)

N <- 1e4
df <- data.frame(a = rnorm(N),
                 b = runif(N),
[...]

PR #486, opened 2022-09-13
Explicit predict instead of do.call

maellecoursonnais:
Following #485, here's an attempt to avoid using `do.call`s with `fixest`. Basically replacing the do.call with the explicit list of arguments, minus the `level` one in the second `do.call`. Not sure whether this is what you were thinking @vincentarelbundock. Tested with the workflow mentioned [here](https://github.com/vincentarelbundock/marginaleffects/issues/485#issuecomment-1245327633), returning:
```r
test-pkg-fixest.R.............    0 tests    0.3s [Exited at #3: problems testing fixest because of `get_data` from environment]
All ok, 0 results (0.3s)
```
Tested with the `N = 1e6` and it is indeed way faster.

### DIFF
--- a/R/get_vcov.R
+++ b/R/get_vcov.R
@@ -82,7 +82,9 @@ get_vcov.default <- function(model,
             "Unable to extract a variance-covariance matrix using this `vcov`
             argument. Standard errors are computed using the default variance
             instead. Perhaps the model or argument is not supported by the
-            `sandwich` package.")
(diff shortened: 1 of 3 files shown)


vincentarelbundock:
Thanks a lot for this, I really appreciate it.

Merged.

Looks like things are several orders of magnitude faster now.

PR #543, opened 2022-11-17
Small speedup

etiennebacher:
Hello Vincent,

`do.call()` + `paste()` is generally faster than `apply()` + `paste()` (that is used in `comparisons()`). It only creates a marginal gain but better than nothing. I used some of the benchmarks of the [vignette on performance](https://vincentarelbundock.github.io/marginaleffects/articles/performance.html). 

### Current (`apply`)

``` r
library(marginaleffects)

# simulate data and fit a large model
N <- 1e5
dat <- data.frame(matrix(rnorm(N * 26), ncol = 26))
mod <- lm(X1 ~ ., dat)

bench::mark(
  marginaleffects(mod, vcov = FALSE, variables = "X3"),
  marginaleffects(mod, variables = "X3"),
  iterations = 10,
  check = FALSE
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 2 × 6
#>   expression                                                min   median itr/s…¹
#>   <bch:expr>                                           <bch:tm> <bch:tm>   <dbl>
#> 1 marginaleffects(mod, vcov = FALSE, variables = "X3")    1.05s    1.16s   0.866
#> 2 marginaleffects(mod, variables = "X3")                  5.63s    6.69s   0.152
#> # … with 2 more variables: mem_alloc <bch:byt>, `gc/sec` <dbl>, and abbreviated
#> #   variable name ¹​`itr/sec`
```

### New (`do.call`)

``` r
library(marginaleffects)

# simulate data and fit a large model
N <- 1e5
dat <- data.frame(matrix(rnorm(N * 26), ncol = 26))
mod <- lm(X1 ~ ., dat)

bench::mark(
  marginaleffects(mod, vcov = FALSE, variables = "X3"),
  marginaleffects(mod, variables = "X3"),
  iterations = 10,
  check = FALSE
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 2 × 6
#>   expression                                                min   median itr/s…¹
#>   <bch:expr>                                           <bch:tm> <bch:tm>   <dbl>
#> 1 marginaleffects(mod, vcov = FALSE, variables = "X3") 615.74ms 674.88ms   1.48 
#> 2 marginaleffects(mod, variables = "X3")                  5.37s    5.94s   0.169
#> # … with 2 more variables: mem_alloc <bch:byt>, `gc/sec` <dbl>, and abbreviated
#> #   variable name ¹​`itr/sec`
```

--- 

Regarding tests, a lot of them didn't run because `ON_CRAN` was true (which obviously wasn't). I think the problem comes from `inst/tinytest/helpers.R`:
https://github.com/vincentarelbundock/marginaleffects/blob/9b5dfdaa28c13bbe74a87745bedf29cd7bd5fa0b/inst/tinytest/helpers.R#L16

~~Shouldn't you remove the `!` at the beginning?~~ Actually it was because `Sys.getenv("R_NOT_CRAN")` return `""` in my case

(Also I have no idea why some spaces were removed on other lines)

### DIFF
--- a/R/comparisons.R
+++ b/R/comparisons.R
@@ -112,7 +112,7 @@
 #' # Contrasts at user-specified values
 #' comparisons(mod, newdata = datagrid(am = 0, gear = tmp$gear))
 #' comparisons(mod, newdata = datagrid(am = unique, gear = max))
-#' 
+#'
 #' m <- lm(mpg ~ hp + drat + factor(cyl) + factor(am), data = mtcars)
 #' comparisons(m, variables = "hp", newdata = datagrid(FUN_factor = unique, FUN_numeric = median))
 #'
@@ -125,7 +125,7 @@
 #' comparisons(mod, variables = list(hp = "iqr")) %>% tidy()
 #' comparisons(mod, variables = list(hp = "sd")) %>% tidy()
 #' comparisons(mod, variables = list(hp = "minmax")) %>% tidy()
-#' 
+#'
 #' # using a function to specify a custom difference in one regressor
 #' dat <- mtcars
 #' dat$new_hp <- 49 * (dat$hp - min(dat$hp)) / (max(dat$hp) - min(dat$hp)) + 1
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
Oh nice! A 5x improvement in some models is non-trivial, and I didn't know that trick, so thanks a lot.

`R_NOT_CRAN` is an environment variable I use to skip some tests. You can set it in your `bash` profile, or by calling this inside R: `Sys.getenv("R_NOT_CRAN")`

vincentarelbundock:
Ah no, I see that the 5x is because of `vcov=FALSE`. Still cool. Thanks.

PR #621, opened 2023-01-31
Added fwb to inferences()

ngreifer:
One thing that might need some attention is how the weights are supplied to the `wts` argument, especially if `marginal_means()` is the function called. Might want to add a check and do some argument processing. Also, `verbose` is set to `TRUE` by default in `fwb()`, so if you want to set that to `FALSE` by default you can to prevent the progress bar from showing.


vincentarelbundock:
This looks fantastic, thanks!

I think I don't want to support bootstrap for `maginal_means()` right now anyway. I added an informative error in those cases.

How confident are you in the weights issue for other object types? I got a revdep email from CRAN so need t prepare a new release in the next few days. Should we take some time to think and wait for the next release before merging?

vincentarelbundock:
Also, does this fail gracefully for model-fitting functions that do not have a `weights` argument, or will it produce errors (or worse: incorrect results)?

ngreifer:
If the user specified their own argument to `wts`, that will be overridden. If they specified the weights as original weights supplied to the model (e.g., IPWs), those will be multiplied by the bootstrap weights and be correctly incorporated. So I think it should be fine. 97% confident it will work as intended unless the `wts` are not those supplied to the original model.

I don't know of any models that don't accept weights, so I don't know how it will fail. What the user will see is that the standard error is equal to 0 because all estimates will be the same. The infrastructure I used here is the same one used in `sandwich`'s implementation of the method. It could probably do with some testing beyond what I've done (just the basics), but I think it could be ready for CRAN. Since this is experimental anyway I think you can get away with leaving it.

Unless you're ready to settle on the interface for `inferences()`, I would not include it in the CRAN update if it is a minor update. I feel that experimental interfaces should be reserved for development until they are finalized. Do what you have to do though.

vincentarelbundock:
That all sounds great.

I definitely agree with you in principle on experimental features. The only issue is that I don't personally bootstrap a lot on a day to day basis, so this will probably just sit there unused, which doesn't help refinement. Marking it as experimental and getting it out might be the only way to get feedback.

I use the other stuff in the package a lot, so it was easier to iterate before public release...

vincentarelbundock:
This is all fantastic. I merged with an extra informative error on `wts` and a very short example in the vignette.

Thanks again for a great feature!

PR #632, opened 2023-02-02
Improve the performance of `get_predict()`

etiennebacher:
Two things:

1. Creating a `data.table()` takes some time and is not necessary in `get_predict()` because there’s no `data.table`-specific operations. It is faster to create a list, which is also accepted by `setDF()`. 
2. If I'm not mistaken, the only reason why you use `as.numeric(pred)` is because of `svyglm` which returns an object of class `svystat`. It's faster to directly assign the class "numeric" when there are a lot of observations.

These improve the performance mostly when `vcov = TRUE`.

I use the same setup as in the [performance vignette](https://vincentarelbundock.github.io/marginaleffects/articles/performance.html) but with less observations because it takes quite some time to run all these benchmarks.

### Before

``` r
library(marginaleffects)

# simulate data and fit a large model
N <- 3*1e4
dat <- data.frame(matrix(rnorm(N * 26), ncol = 26))
mod <- lm(X1 ~ ., dat)

results <- bench::mark(
  # marginal effects at the mean; no standard error
  slopes(mod, vcov = FALSE, newdata = "mean"),
  # marginal effects at the mean
  slopes(mod, newdata = datagrid()),
  # 1 variable; no standard error
  slopes(mod, vcov = FALSE, variables = "X3"),
  # 1 variable
  slopes(mod, variables = "X3"),
  # 26 variables; no standard error
  slopes(mod, vcov = FALSE),
  # 26 variables
  slopes(mod),
  iterations = 10,
  check = FALSE
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.

results[, c(1, 3, 5)]
#> # A tibble: 6 × 3
#>   expression                                    median mem_alloc
#>   <bch:expr>                                  <bch:tm> <bch:byt>
#> 1 slopes(mod, vcov = FALSE, newdata = "mean") 419.16ms  415.82MB
#> 2 slopes(mod, newdata = datagrid())           668.29ms  421.67MB
#> 3 slopes(mod, vcov = FALSE, variables = "X3") 305.74ms  212.78MB
#> 4 slopes(mod, variables = "X3")                  2.06s    1.23GB
#> 5 slopes(mod, vcov = FALSE)                      6.71s    4.13GB
#> 6 slopes(mod)                                   41.76s   28.59GB
```

### After

``` r
results[, c(1, 3, 5)]
#> # A tibble: 6 × 3
#>   expression                                    median mem_alloc
#>   <bch:expr>                                  <bch:tm> <bch:byt>
[...]

PR #635, opened 2023-02-03
Additional performance gains

### LINKED ISSUE #628: Some potential speed improvements
I noticed a few places in the code where I think the speed could be improved. If you think it's worth implementing and you don't have time for this, I can maybe make a PR.

### Delete/select columns by reference

Here for example:
https://github.com/vincentarelbundock/marginaleffects/blob/11d66dabd7c0636051a65c6ba0589629f1258d3c/R/comparisons.R#L463

``` r
library(data.table)

# Make a big data.table
tmp <- vector("list", length = 1e6)
for (i in 1:1e6) {
  tmp[[i]] <- mtcars
}
test <- rbindlist(tmp)

cols = names(test)[1:2]

bench::mark(
  assign = {
    test <- test[, ..cols, drop = FALSE]
    head(test)
  },
  ref = {
    test[, setdiff(names(test), cols) := NULL]
    head(test)
  }
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 assign        152ms    152ms      6.58     488MB     6.58
#> 2 ref           286µs    335µs   2405.       199KB    10.5
```

But need to be careful if data is modified by reference

### Use `collapse::fduplicated()`

Only available in devel version of `collapse` for now (and `duplicated()` used only once in the code):

``` r
set.seed(123)
test <- sample(1:10, 1e7, replace = TRUE)

bench::mark(
  duplicated(test),
  collapse::fduplicated(test)
)
#> disabled.
#> # A tibble: 2 × 6
#>   expression                       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 duplicated(test)              87.7ms   95.4ms      9.85     166MB     9.85
#> 2 collapse::fduplicated(test)   21.1ms   24.1ms     33.0       86MB    11.6
```
I hope it's useful

---

I realized afterwards that `collapse` is suggested only, not imported by `marginaleffects` so it would add a dep

etiennebacher:
Some additional performance gains:

* `pred_lo` is a data.table but not `pred_hi` so extracting values with `[[` took much more time (and is called a lot by `slopes`). That's why we convert `pred_hi` to a data.table (I can't reproduce this with local benchmarks but it makes quite a difference).
* `data.table`'s method for `na.omit()` is faster than subsetting with `is.na()`
* selecting columns by reference, as shown in #628 

Notes: need to be careful about operations by reference because they can have side effects by modifying the original dataset. I relied only on the tests suite here, I didn't check for additional cases.

### Benchmarks

(To compare with the results in #632)

``` r
library(marginaleffects)
N <- 3*1e4
dat <- data.frame(matrix(rnorm(N * 26), ncol = 26))
mod <- lm(X1 ~ ., dat)

results <- bench::mark(
  slopes(mod, vcov = FALSE),
  slopes(mod),
  iterations = 10,
  check = FALSE
)
#> disabled.

results[, c(1, 3, 5)]
#> # A tibble: 2 × 3
#>   expression                  median mem_alloc
#>   <bch:expr>                <bch:tm> <bch:byt>
#> 1 slopes(mod, vcov = FALSE)     4.4s    3.34GB
#> 2 slopes(mod)                  19.3s   23.57GB
```

Close #628


R/get_contrasts.R
```diff
@@ -288,7 +290,7 @@ get_contrasts <- function(model,
 
     # drop missing otherwise get_averages() fails when trying to take a simple mean
     idx_na <- !is.na(out$predicted_lo)
```
vincentarelbundock, line 292: Cool! We could also move `na.omit()` after the bayes stuff and only compute `idx_na` if we need to operate on `draws`.

vincentarelbundock: I guess this would have to be a "true" if-else, where `na.omit()` is also called in the bayes block because `out` is used there too.

etiennebacher: But I think `out` without the `NA`s is used a few lines after in `subset(out,...)`? So I guess we still need to use `na.omit()` before the bayes stuff

etiennebacher: Maybe I'm confusing several things, you can also commit directly to the branch if you prefer

etiennebacher: Done

PR #680, opened 2023-02-17
Don't use R 4.0 pipe in code

strengejacke:
_marginaleffects_ depends on R > 3.5, so `|>` does not work.


vincentarelbundock:
Ooof, that's a silly one. Thanks!

vincentarelbundock:
But I need to test if that unique is even necessary. I think they might be some prototyping code I left in by accident.

PR #706, opened 2023-03-06
Small speedup for matrix multiplication

etiennebacher:
From [this SO answer](https://stackoverflow.com/questions/35923787/fast-large-matrix-multiplication-in-r):
``` r
n_cols = 30

my_J = matrix(sample(1:50, n_cols * 2500000, TRUE), ncol = n_cols)
my_V = matrix(sample(1:10, n_cols^2, TRUE), ncol = n_cols)

bench::mark(
  my_J %*% my_V,
  crossprod(t(my_J), my_V),
  iterations = 10
)
#> # A tibble: 2 × 6
#>   expression                    min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>               <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 my_J %*% my_V               3.35s    3.35s     0.299    1.12GB     2.99
#> 2 crossprod(t(my_J), my_V)    2.08s    2.08s     0.481     1.4GB     5.29
```

Doesn't really make sense to benchmark with `marginaleffects` functions since it's a small change in the total time

---

Besides this, I'm seeing `data.table` warnings in the tests (even before this change) but I don't know where they come from.
```
1: In data.table::setDT(modeldata) :
  Some columns are a multi-column type (such as a matrix column): [3]. setDT will retain 
these columns as-is but subsequent operations like grouping and joining may fail. Please
 consider as.data.table() instead which will create a new column for each embedded column.
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: marginaleffects
 Title: Predictions, Comparisons, Slopes, Marginal Means, and Hypothesis Tests
-Version: 0.10.0.9012
+Version: 0.10.0.9013
 Authors@R: 
     c(person(given = "Vincent",
              family = "Arel-Bundock",
--- a/R/comparisons.R
+++ b/R/comparisons.R
@@ -271,15 +271,15 @@ comparisons <- function(model,
         df = df),
         list(...))
     call_attr <- do.call("call", call_attr)
-    
+
     # multiple imputation
     if (inherits(model, "mira")) {
         out <- process_imputation(model, call_attr)
         return(out)
     }
-    
+
     # more sanity chekcs
-    
+
     conf_level <- sanitize_conf_level(conf_level, ...)
     sanity_dots(model, ...)
     checkmate::assert_number(eps, lower = 1e-10, null.ok = TRUE)
@@ -310,7 +310,7 @@ comparisons <- function(model,
     } else {
         addvar <- FALSE
     }
-    
+
     # extracting modeldata repeatedly is slow.
     # checking dots allows cheap multiple imputation
     dots <- list(...)
@@ -326,13 +326,13 @@ comparisons <- function(model,
         }
         modeldata <- get_modeldata(model, additional_variables = addvar)
     }
-    
+
     newdata <- sanitize_newdata(
         model = model,
         newdata = newdata,
         by = by,
         modeldata = modeldata)
-    
+
     # weights: before sanitize_variables
     sanity_wts(wts, newdata) # after sanity_newdata
     if (!is.null(wts) && isTRUE(checkmate::check_string(wts))) {
@@ -465,7 +465,7 @@ comparisons <- function(model,
         if ("rowid" %in% colnames(newdata)) {
             idx <- c("rowid", "rowidcf", "term", "contrast", "by", setdiff(colnames(contrast_data$original), colnames(mfx)))
             idx <- intersect(idx, colnames(contrast_data$original))
-            tmp <- contrast_data$original[, ..idx, drop = FALSE]
+            tmp <- contrast_data$original[, ..idx]
             # contrast_data is duplicated to compute contrasts for different terms or pairs
             bycols <- intersect(colnames(tmp), colnames(mfx))
             idx <- duplicated(tmp, by = bycols)
--- a/R/get_se_delta.R
+++ b/R/get_se_delta.R
@@ -114,7 +114,7 @@ get_se_delta <- function(model,
     # Var(dydx) = J Var(beta) J'
     # computing the full matrix is memory-expensive, and we only need the diagonal
     # algebra trick: https://stackoverflow.com/a/42569902/342331
-    se <- sqrt(colSums(t(J %*% V) * t(J)))
+    se <- sqrt(colSums(t(crossprod(t(J), V)) * t(J)))
     se[se == 0] <- NA_real_
     attr(se, "jacobian") <- J
 


vincentarelbundock:
This is interesting. I wonder what the priority should be here: speed or memory. See last line which is a bit more readable and more memory-efficient:

``` r
n_cols = 30
J = matrix(rnorm(n_cols * 2500000), ncol = n_cols)
V = matrix(runif(n_cols^2), ncol = n_cols)

bench::mark(
    colSums(t(J %*% V) * t(J)),
    colSums(t(crossprod(t(J), V)) * t(J)),
    rowSums(tcrossprod(J, V) * J),
    iterations = 5
)

# # A tibble: 3 × 6
#   expression                                 min   median `itr/sec` mem_alloc `gc/sec`
#   <bch:expr>                            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
# 1 colSums(t(J %*% V) * t(J))               2.88s    3.17s     0.313     1.7GB    0.313
# 2 colSums(t(crossprod(t(J), V)) * t(J))    2.68s     3.1s     0.326    2.25GB    0.521
# 3 rowSums(tcrossprod(J, V) * J)            2.99s    3.34s     0.302    1.14GB    0.181
```

The warning is probably in the `mlogit` or `mblogit` tests. Could diagnose by setting `options(warn=2)` before testing. My guess is that this is not an important warning and we chould just suppress the warning around that `setDT` call.

vincentarelbundock:
I'm asking because this operation is only executed once, and the difference is a fraction of a second for massive datasets.

etiennebacher:
Interesting, you have very small time differences while in my case the gap is around 30%. But in any case this is indeed a small change and it consumes more memory. Maybe it's not worth changing it.

FYI, I think I'll stop the optimization here because it seems 90-95% of the time spent is now in `get_predict()` in all `marginaleffects` functions (except maybe for bayesian models, I didn't look at them at all).

Just to provide a complete benchmark if someone finds it useful:
```r
bench::press(
  n_cols = c(1, 10, 20, 30, 40),
  {
    J = create_J(n_cols)
    V = create_V(n_cols)
    bench::mark(
      colSums(t(J %*% V) * t(J)),
      colSums(t(crossprod(t(J), V)) * t(J)),
      rowSums(tcrossprod(J, V) * J),
      iterations = 5
    )
  }
)

Running with:
  n_cols
1      1
2     10
3     20
4     30
5     40
# A tibble: 15 × 14
   expression                            n_cols      min   median `itr/sec` mem_al…¹ gc/se…² n_itr  n_gc total_…³ result memory     time      
   <bch:expr>                             <dbl> <bch:tm> <bch:tm>     <dbl> <bch:by>   <dbl> <int> <dbl> <bch:tm> <list> <list>     <list>    
 1 colSums(t(J %*% V) * t(J))                 1  30.41ms  34.97ms    28.8    76.29MB   0         5     0 173.42ms <dbl>  <Rprofmem> <bench_tm>
 2 colSums(t(crossprod(t(J), V)) * t(J))      1  32.41ms  33.92ms    26.1    95.37MB   0         5     0 191.67ms <dbl>  <Rprofmem> <bench_tm>
 3 rowSums(tcrossprod(J, V) * J)              1   34.8ms  40.27ms    24.8    57.22MB   0         5     0 201.28ms <dbl>  <Rprofmem> <bench_tm>
 4 colSums(t(J %*% V) * t(J))                10 512.02ms 530.88ms     1.82  591.28MB   0.456     4     1    2.19s <dbl>  <Rprofmem> <bench_tm>
 5 colSums(t(crossprod(t(J), V)) * t(J))     10 472.36ms 472.36ms     2.12  782.01MB   8.47      1     4 472.36ms <dbl>  <Rprofmem> <bench_tm>
 6 rowSums(tcrossprod(J, V) * J)             10 480.84ms 521.93ms     1.93  400.54MB   0.484     4     1    2.07s <dbl>  <Rprofmem> <bench_tm>
 7 colSums(t(J %*% V) * t(J))                20    1.77s    1.83s     0.553   1.14GB   0.553     5     5    9.04s <dbl>  <Rprofmem> <bench_tm>
 8 colSums(t(crossprod(t(J), V)) * t(J))     20    1.28s    1.43s     0.711   1.51GB   0.996     5     7    7.03s <dbl>  <Rprofmem> <bench_tm>
 9 rowSums(tcrossprod(J, V) * J)             20    1.56s    1.61s     0.596 782.01MB   0.119     5     1    8.39s <dbl>  <Rprofmem> <bench_tm>
10 colSums(t(J %*% V) * t(J))                30    3.57s    3.63s     0.271    1.7GB   0.379     5     7   18.47s <dbl>  <Rprofmem> <bench_tm>
11 colSums(t(crossprod(t(J), V)) * t(J))     30    2.61s    2.72s     0.369   2.25GB   0.369     5     5   13.55s <dbl>  <Rprofmem> <bench_tm>
12 rowSums(tcrossprod(J, V) * J)             30    3.29s    3.59s     0.284   1.14GB   0.171     5     3   17.58s <dbl>  <Rprofmem> <bench_tm>
13 colSums(t(J %*% V) * t(J))                40    5.98s    6.12s     0.164   2.25GB   0.262     5     8   30.56s <dbl>  <Rprofmem> <bench_tm>
14 colSums(t(crossprod(t(J), V)) * t(J))     40    4.26s    4.43s     0.225      3GB   0.496     5    11    22.2s <dbl>  <Rprofmem> <bench_tm>
15 rowSums(tcrossprod(J, V) * J)             40    5.66s    5.81s     0.172   1.51GB   0.138     5     4   29.03s <dbl>  <Rprofmem> <bench_tm>
```

vincentarelbundock:
Thanks for those. The variation is quite interesting.

> FYI, I think I'll stop the optimization here because it seems 90-95% of the time spent is now in get_predict() in all marginaleffects functions

Yeah, that makes sense. Thanks so much for all of those. I think in the last two versions we probably got speed-ups of 6-8x for simple GLM models. (Probably a bit less in other packages, because their `predict` methods tend to be less optimized, and we spend even more time in `get_predict()`). So this is very very cool.

For `lm()` models we could probably get a 2x speedup by building our own design matrix and multiplying it, but that would probably be a bit less safe. Overall, rolling out our own faster `predict()` methods feels like a bit of a rabbit hole (and also dangerous), so perhaps that is the end of optimization for now.

PR #713, opened 2023-03-08
Performance: Handcode `get_predict()` w/ `RcppEigen`

vincentarelbundock:
Source of inefficiency:

* With `vcov=TRUE`, we must execute `predict()` twice for each coefficient in the model.
* When we call `predict()`, it builds a whole new `model.matrix()` on every call.
* But the model matrix does not usually change from call to call, so this is wasteful.

Solution:

* `get_model_matrix_attributes()` builds a model matrix and attaches it as an attribute to `newdata`.
* `get_model_matrix_attributes()` is conservative in returning `NULL` in many unsupported cases, such as `NA` or `offset`
* When `get_predict()` sees a model matrix attribute and a "standard" case, it uses `RcppEigen` to do compute the predictions manually.

This gives us about 4x speedups in non-crazy scenarios (0.5x at the low end, and 10x in some cases). Also, memory use is much improved (2-10x reduction).

This PR only supports `lm()` and `glm()`, but it would be reasonably easy to cover a lot of the models. Mixed effects and GAM are probably out of scope because their `predict()` methods are very complex, so they will likely remain slow. But a ton of popular models seem to have very simple `predict()` methods, such as `qr`, `MASS::polr`, etc.

See attached files for comparisons to 0.10.0 using the cases in the PR version of `sandbox/benchmarks.Rmd`.

[benchmarks_reprex_0_10_0.md](https://github.com/vincentarelbundock/marginaleffects/files/10922889/benchmarks_reprex_0_10_0.md)
[benchmarks_reprex.md](https://github.com/vincentarelbundock/marginaleffects/files/10922888/benchmarks_reprex.md)


vincentarelbundock:
This PR might be of interest to @etiennebacher

etiennebacher:
Just a few quick thoughts:

1. About results and deps

* benchmarks improvements are very impressive!
* I didn't understand which part led to this large improvement: is it the fact that the model.matrix is now called only once instead of being repeated, or is it the use of `RcppEigen` (or both)?
* Never used `RcppEigen` so can't tell much about it. I actually thought that using `Rcpp` required some knowledge of C / C++ but it seems that all of that is taken care of automatically
* Adds 4 dependencies (one direct, `RcppEigen`, 3 indirect, `Rcpp`, `lattice`, `Matrix`)

2. About model support

* better if package is fast for sure
* adds a lot of maintenance burden if you need to take care of the predict methods in addition to all the rest
* is it better to support few models fully (including offset or special cases) or a lot of models partially?
* maybe a bit naive but you could do a twitter poll to know which family of models people use the most? this could give you some hints on which predict methods should have the priority if they are not too complex (but ofc the sample is probably not representative)

I don't have a clear-cut opinion on these points (and even if I did, you're the package author and maintainer so who cares what I think ^^), I just wanted to mention them in case it's helpful for you.

vincentarelbundock:
>     * I didn't understand which part led to this large improvement: is it the fact that the model.matrix is now called only once instead of being repeated, or is it the use of `RcppEigen` (or both)?

It's clearly both. I'm trying to figure out the relative contribution now, but this is kind of tricky to benchmark.

>     * Never used `RcppEigen` so can't tell much about it. I actually thought that using `Rcpp` required some knowledge of C / C++ but it seems that all of that is taken care of automatically

Yeah, this turned out to be super easy. It was just ~4 lines of C++.

>     * Adds 4 dependencies (one direct, `RcppEigen`, 3 indirect, `Rcpp`, `lattice`, `Matrix`)

I believe that `lattice` and `Matrix` are installed by default with `R`, so this is really about `RcppEigen` and `Rcpp`, and those are rock solid deps, maintained by Dirk.

>     * adds a lot of maintenance burden if you need to take care of the predict methods in addition to all the rest

Yes, this is the main issue that I need to consider.

>     * is it better to support few models fully (including offset or special cases) or a lot of models partially?

I think it's pretty easy to do partial support and fall back on the default `predict()` approach, but would have to see on a case-by-case basis.

Will think about this, but right now it feels like this makes sense, at least for the very common `lm` and `glm` cases.

Issue #761, opened 2023-04-28
 inferences() fails with large data sets


amcmahon17:
Hello,

I haven't gotten inferences() to complete on large datasets (100k or above), even with n=10; sometimes I get the following error: invalid type (closure) for variable '(weights)'.

Here's some example code.

```r
library(marginaleffects)
library(tidyverse)
data(mtcars)
big_mtcars<-mtcars %>% sample_n(100000,replace=TRUE)

#unclear how long this takes

system.time(
  avg_predictions(lm(mpg~.,data=big_mtcars)) %>% inferences(method="rsample")
)


#for comparison, completes in several minutes

boots <- bootstraps(big_mtcars, times = 1000)
            
fit_lm_on_bootstrap <- function(split) {
  lm(mpg ~., data= analysis(split))
}

boot_models <-
boots %>% 
mutate(model = map(splits, fit_lm_on_bootstrap),
         coef_info = map(model, tidy))

int_pctl(boot_models, coef_info)
```

vincentarelbundock:
Thanks a lot for the report and the clear reproducible example. I see the same problem on my computer. I am traveling and overcommitted with "real work", so i can't look into this seriously now, but I'll ping you if/when I find a solution.

amcmahon17:
No problem.

I do want to compliment you on the package. I get quite a bit out of it and
I'm impressed by how easy it is to use.

PR #771, opened 2023-05-09
Large speed improvement for `inferences()`

etiennebacher:
Related to #761. The bootstrap is slow because it calls `get_modeldata()` a lot inside `predictions()`, which is `FUN` in these lines:

https://github.com/vincentarelbundock/marginaleffects/blob/274d9195ecc48a3430de76562187c0ddbe89687c/R/bootstrap_rsample.R#L34-L44 

What I'm thinking is that you already know the model data, it's the line `modcall[["data"]] <- data`. So you could pass this data directly to `predictions()` and avoid running `get_modeldata()`. 

This leads to very large performance gains:

``` r
library(dplyr, warn.conflicts = FALSE)
library(marginaleffects)

set.seed(1234)
big_mtcars <- mtcars %>% 
  sample_n(15000, replace = TRUE)

mod <- lm(mpg ~ .,data=big_mtcars)
foo <- avg_predictions(mod) 

### Before
bench::mark(
  test = inferences(foo, method="rsample", R = 30)
)
#> Warning: Recommend at least 1000 non-missing bootstrap resamples for term `1`.
#> Recommend at least 1000 non-missing bootstrap resamples for term `1`.
#> disabled.
#> # A tibble: 1 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 test          33.2s    33.2s    0.0301    5.25GB     1.02

### After
bench::mark(
  test = inferences(foo, method="rsample", R = 30)
)
#> Warning: Recommend at least 1000 non-missing bootstrap resamples for term `1`.
#> Recommend at least 1000 non-missing bootstrap resamples for term `1`.
#> disabled.
#> # A tibble: 1 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 test          6.48s    6.48s     0.154     830MB     1.54
```

However, there are no numerical tests (only checks for class of `inferences()` output). I did a few quick checks but these are not comprehensive.

Also, I was limited in time so I just made this for `rsample` but it should be similar for the others (+ I'd like to be sure this is fine before I spend time on the rest).

It is still surprising that it takes several seconds for just 30 reps and 15,000 obs.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: marginaleffects
 Title: Predictions, Comparisons, Slopes, Marginal Means, and Hypothesis Tests
-Version: 0.11.1.9012
+Version: 0.11.1.9013
 Authors@R: 
     c(person(given = "Vincent",
              family = "Arel-Bundock",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,14 @@
-# marginaleffects 0.11.1.9001
+# marginaleffects 0.11.1.9003
 
 * Strings in `wts` are accepted with the `by` argument.
 * `predictions()` and `avg_predictions()` no longer use an automatic backtransformation for GLM models unless `hypothesis` is `NULL`.
 * `vcov()` can be used to retrieve a full variance-covariance matrix from objects produced by `comparisons()`, `slopes()`, `predictions()`, or `marginal_means()` objects.
 * When processing objects obtained using `mice` multiple imputation, the pooled model using `mice::pool` is attached to the `model` attribute of the output. This means that functions like `modelsummary::modelsummary()` will not erroneously report goodness-of-fit statistics from just a single model and will instead appropriately report the statistics for the pooled model. Thanks to @Tristan-Siegfried for PR #740.
 * More informative error messages on some prediction problems. Thanks to @andymilne for Report #751.
+
+Performance:
+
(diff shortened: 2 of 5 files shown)


vincentarelbundock:
Great catch!

I'm flying today but I'll look at this when I find a minute and try to implement the other engines.

etiennebacher:
Just for reference, updating this branch with #770 leads to even better changes:
``` r
bench::mark(
  test = inferences(foo, method="rsample", R = 30)
)
#> Warning: Recommend at least 1000 non-missing bootstrap resamples for term `1`.
#> Recommend at least 1000 non-missing bootstrap resamples for term `1`.
#> disabled.
#> # A tibble: 1 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 test          1.87s    1.87s     0.535     815MB     6.42
```

etiennebacher:
I added a few numerical checks and the speedup for method "boot". For "fwb" I didn't see any speedup at all so I didn't change anything for this

vincentarelbundock:
This is awesome! I really appreciate that you took the time to do this!

Issue #772, opened 2023-05-11
Ambiguity about specifying values versus increments in comparisons vignette


bgall:
While the [comparisons vignette](https://vincentarelbundock.github.io/marginaleffects/reference/comparisons.html), mostly provides example with factor or string predictors, there are a couple of examples of [numeric predictors](https://vincentarelbundock.github.io/marginaleffects/articles/comparisons.html#numeric-predictors). One example explains that: 

> ...we can see what happens to the adjusted predictions when we increment the hp variable by...5 units:
>
>` avg_comparisons(mod, variables = list(hp = 5))`

The subsequent [subsection on differences](https://vincentarelbundock.github.io/marginaleffects/articles/comparisons.html#differences) explains that under the hood the predictions above are performed by calculating the difference in model predictions for `hp - 2.5` and `hp + 2.5` and we can always change the centering as we please.

What is unclear in the vignette, however, is how behavior changes when >1 _numeric_ value for a variable is passed to the `variables` argument. For example, suppose we had instead used: `avg_comparisons(mod, variables = list(hp = c(1,3))`.

My expectation would be that this make comparisons between predictions when `hp = 1` and when `hp = 3` rather than between incrementing `hp` by 1 and incrementing `hp` by 3. In other words, it simply sets the vector to the specified values. Is this accurate? If so, it would be helpful to explicate in the vignette that the "differences" behavior changes when more than one value is used.

vincentarelbundock:
I think the documentation f the argument is pretty clear on this point. Do you agree? If not, can you suggest a clearer wording? See: https://vincentarelbundock.github.io/marginaleffects/reference/comparisons.html

I'd be *very* happy to merge a pull request if you would like to improve the vignette.

bgall:
Happy to submit a PR clarifying, but wanted to verify my understanding before editing.

vincentarelbundock:
Cool, thanks. A vector of two definitely compares the specific two values.

PR #774, opened 2023-05-11
Update vignette to explicate how comparison functions treat vectors of length 1 versus >1

bgall:
How vectors of length 1 differ from those of length >1 is fairly well-described in the function reference but was somewhat unclear in the vignette. I clarify here since it can be easy to miss otherwise.


bgall:
Addresses issue https://github.com/vincentarelbundock/marginaleffects/issues/772

vincentarelbundock:
These are excellent edits. Thanks a lot, I really appreciate it!

PR #781, opened 2023-05-16
Support for flexsurv models

mattwarkentin:
Hi @vincentarelbundock,

This is some ground work to support `{flexsurv}` models of class `flexsurvreg`. I have marked this as a draft PR while I seek your feedback and iterate.


vincentarelbundock:
Awesome!

A couple comments written while commuting:

Do we need a `sanitize_model_specific()` method to return an informative error if the user does not supply `times`? See example here: https://github.com/vincentarelbundock/marginaleffects/blob/main/R/methods_brms.R

Add class to sanity_model.R

Add class to data-raw/supported_models.csv with TRUE under B and SE, but empty cells to indicate that we did not compare to other softwares. 

A couple simple tests if possible. Do you know if we can compare to benchmark estimates from alternative software?

mattwarkentin:
> Do we need a sanitize_model_specific() method to return an informative error if the user does not supply times?

We may. I made a decision consideration to force the requirement for `times` and also to force it to be a scalar. By default, `flexsurv` makes predictions at all observed event times or the user can supply a vector of times for predictions. It would then return a `tibble` with a list-column of predictions. This seemed to go against the design/expectation of `marginaleffects` so I thought the simplest approach was to only allow predictions at a single fixed horizon. Do you agree?

I will get to work on your other comments.

mattwarkentin:
> A couple simple tests if possible. Do you know if we can compare to benchmark estimates from alternative software?

We may be able to benchmark against the `stpm2` implementation of flexible parametric survival models in Stata - I haven't used this in a while but I can look into.

mattwarkentin:
I agree that users will likely want to predict at several time horizons. Glad to see this type of multi-prediction is already supported by this package. This would actually allow us to remove the need to specify `times` as required because it is optional for `predict.flexsurvreg()`.

My only current sticking point is that in some cases we need to `unnest()` the list-column returned by `predict.flexsurvreg()` but I suppose you don't want to add `{tidyr}` as a dependency. Not sure of a safe workaround yet.

vincentarelbundock:
Yeah, I don't want to add new full dependencies. If the format of the list column is consistent and predictable, it's usually pretty easy to unnest manually using base R. But if you really can't figure it out, the last recourse could be to add `tidyr` to Suggests in DECRIPTION, and to call `insight::check_if_installed("tidyr")` in `sanity_model_specific.flexsurv`

mattwarkentin:
The list-columns are predictable, and I have now solved that issue. I think I may need to add some info to `type_dictionary_build()` to allow support the various prediction types. Right?

mattwarkentin:
I get this warning when trying to pass the `times` argument into `predictions()`:
>Warning message:
These arguments are not supported for models of class `flexsurvreg`: times, se.fit. Please file a request on Github if you believe that additional arguments should be supported: https://github.com/vincentarelbundock/marginaleffects/issues

vincentarelbundock:
Ah yes, sorry, forgot about those.

The order of rows in `type_dictionary` determines the default value, so it matters a lot.

The warning can be silenced by adding an appropriate  `times` entry in R/sanity_dots.R

mattwarkentin:
I've made those changes. Getting closer.

After adding valid arguments to `sanity_dots.R`, I am now running into the following issue with what seems like partial arg matching between the `p` argument to `predict.flexsurvreg()` and the `p_adjust` argument:
```r
predictions(model, type = 'quantile', p = 0.5)
```
```
Error in get_ci(tmp, conf_level = conf_level, vcov = vcov, draws = draws, :
Assertion on 'p_adjust' failed: Must be element of set {'holm','hochberg','hommel','bonferroni','BH','BY','fdr','none'}, but types do not match (numeric != character).
```

vincentarelbundock:
Ooof, that's a tough one. I don't have a great solution for this one off the top of my head. Might have to try a few things.

I was thinking about reassigning p_adjust from sys.call(), but the p argument might not pass through ...

Last resort might be to use a different argument name altogether and to raise an informative error if users try p.

(Perhaps passing p_adjust to sanity_model_specific to check it.)

vincentarelbundock:
@mattwarkentin I was looking at this today, and realized that this was much closer to ready than I thought. I pushed a few minor changes and merged.

Thanks for all your work! And sorry for the delayed review and changes.

If you ever have time to look at this again, what we are missing is a few tests to check the numerical accuracy of results. I don't know these models. Is there a benchmark implementation in some other software we could compare against?

Issue #828, opened 2023-06-26
glmmTMB models result in error that vcov arguments are not supported.


Dpananos:
When passing models fit using `glmmTMB` an error is raised stating standard errors are not supported for models of class glmmTMB and to set vcov to False.

``` r
library(tidyverse)
library(marginaleffects)
library(glmmTMB)


d <- lme4::sleepstudy

fit <- glmmTMB(Reaction ~ Days + (1|Subject), data=d)

predictions(fit)
#> Error: Standard errors are not supported for models of class `glmmTMB`. Please
#>   set `vcov = FALSE`.

sessionInfo()
#> R version 4.1.3 (2022-03-10)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Ubuntu 18.04.6 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
#> LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.2.20.so
#> 
#> locale:
#>  [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
#>  [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
#>  [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
#>  [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
#>  [9] LC_ADDRESS=C               LC_TELEPHONE=C            
#> [11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#>  [1] glmmTMB_1.1.7          marginaleffects_0.13.0 lubridate_1.9.2       
#>  [4] forcats_1.0.0          stringr_1.5.0          dplyr_1.1.0           
#>  [7] purrr_1.0.1            readr_2.1.4            tidyr_1.3.0           
#> [10] tibble_3.1.8           ggplot2_3.4.1          tidyverse_2.0.0       
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.10         lattice_0.20-45     digest_0.6.31      
#>  [4] utf8_1.2.2          R6_2.5.1            reprex_2.0.2       
[...]

Issue #810, opened 2023-06-08
`glmmTMB`: Standard Errors and `REML=TRUE`


strengejacke:
I noticed this when writing a vignette for ggeffects. The standard errors for glmmTMB predictions of random effects are smaller than what `predict()` returns. Is this intended or a bug?

Reprex requires _datawizard_ dev version...

``` r
library(ggeffects)   # predictions and significance testing
library(insight)     # extracting random effects variances
library(datawizard)  # data wrangling and preparation
library(glmmTMB)     # multilevel modelling

# sample data set
data(efc, package = "ggeffects")

efc <- efc |>
  # numeric to factors, set labels as levels
  to_factor(select = c("c161sex", "c172code", "c175empl")) |>
  # recode age into three groups
  recode_values(
    select = "c160age",
    recode = list(`1` = "min:40", `2` = 41:64, `3` = "65:max")
  ) |>
  # rename variables
  data_rename(
    pattern = c("c161sex", "c160age", "quol_5", "c175empl"),
    replacement = c("gender", "age", "qol", "employed")
  ) |>
  # age into factor, set levels, and change labels for education
  data_modify(age = factor(age, labels = c("-40", "41-64", "65+")))

m_null <- glmmTMB(qol ~ 1 + (1 | gender:employed:age), data = efc)


do.call(cbind, predict(
  m_null,
  newdata = data_grid(m_null, c("gender", "employed", "age")),
  se.fit = TRUE
))
#>                  fit    se.fit
#> eta_predict 15.24091 0.8758925
#> eta_predict 15.17225 0.6872397
#> eta_predict 16.12010 0.7850133
#> eta_predict 16.03567 0.6105341
#> eta_predict 14.93677 0.6589485
#> eta_predict 13.75407 0.3305532
#> eta_predict 15.46836 0.5570139
#> eta_predict 14.09512 0.3461792
#> eta_predict 14.41308 0.6287481
#> eta_predict 13.53274 0.4361047
#> eta_predict 15.34190 1.0851461
#> eta_predict 14.85284 1.0382235

ggpredict(
  m_null,
  c("gender", "employed", "age"),
  type = "random",
  interval = "confidence"
) |> as.data.frame()
#>         x predicted std.error conf.low conf.high group facet
#> 1    Male  15.24091 0.8758925 13.52419  16.95763    no   -40
#> 2    Male  14.93677 0.6589485 13.64525  16.22829    no 41-64
[...]

PR #841, opened 2023-07-13
related to #810 and #828

strengejacke:
@vincentarelbundock I will also add tests later. Feel free to modify or suggest other wording for the error message. I think error is OK because I'm not sure the warning is recognized in 100% of all use cases. But we may switch to a warning, too, however, now that we have the possibility to avoid this error, an error is also ok.

### DIFF
--- a/R/methods_glmmTMB.R
+++ b/R/methods_glmmTMB.R
@@ -63,8 +63,8 @@ set_coef.glmmTMB <- function(model, coefs, ...) {
 sanitize_model_specific.glmmTMB <- function(model, vcov = NULL, calling_function = "marginaleffects", ...) {
     if (isTRUE(vcov) || is.null(vcov)) {
         insight::format_error(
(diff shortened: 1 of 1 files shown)


strengejacke:
I'm not familiar with your CI/test-setup, but it _seems_ like the current errors are unrelated to my commit.

vincentarelbundock:
Thanks, this looks great!

I’m being picky now, but I feel the language might be a bit too accusatory. Frankly, I’m not sure I understand the source of the problem, so maybe best to just link to the issue on the glmmTMB GitHub?

sorry to ask for changes!

strengejacke:
Yes, no problem! I'll revise the wording (though, as Northern German, everything we say sounds accusatory, even compliments, so I may need some tries for the final wording :-)).

vincentarelbundock:
awesome, thanks!

PR #881, opened 2023-08-23
Conformal

vincentarelbundock:
+ [x] `insight::get_data()` might be dangerous in `conformal_cv_plus()` if people modify the dataset
+ [x] score functions only work when `type` is on the same scale as the response
+ [x] Column label % is made out of the `conf_level` in the `predictions()` function. Maybe I should not have a `conf_level` in `inferences()` at all, and always use the one in the original call.
+ [x] PR to awesome conformal prediction in the R section.
+ [x] The purpose of confidence intervals is not to catch predictions. Vignette should explicitly say that this is a common misunderstanding that we want to clarify.
+ [x] Show coverage properties of the confidence interval for mean and predictions.
+ [x] Formal statement of the assumptions and coverage guarantee. It is not guaranteed to be valid in each stratum of X.
+ [x] Example where conformal works badly.
+ [x] Describe the split and CV+ algorithms.
+ [x] conformal inference -> conformal prediction


R/conformal.R
```diff
@@ -0,0 +1,110 @@
...
+    return(out)
+}
+
+
+conformal_cv_plus <- function(x, test, R, score, conf_level, ...) {
+    # cross-validation
+    train <- insight::get_data(attr(x, "model"))
+    idx <- sample(seq_len(nrow(train)), nrow(train))
+    idx <- split(idx, ceiling(seq_along(idx) / (length(idx) / R)))
+    scores <- NULL
```
etiennebacher, line 85: depending on the size of `idx`, you may want to pre-allocate the size of `scores` since you reallocate it in each iteration of the loop (I didn't try the code so not sure `idx` is large enough so that this makes a difference)

R/conformal.R
```diff
@@ -0,0 +1,110 @@
...
+    } else if(score == "softmax") {
+      model <- attr(x, "model")
+      response <- x[[insight::find_response(model)]]
+      if (is.numeric(response) && all(response %in% 0:1)) {
+        # See p.4 of Angelopoulos, Anastasios N., and Stephen Bates. 2022. “A
+        # Gentle Introduction to Conformal Prediction and Distribution-Free
+        # Uncertainty Quantification.” arXiv.
+        # https://doi.org/10.48550/arXiv.2107.07511.
+        # 1 minus the softmax output of the true class
+        out <- 1 - ifelse(response == 1, x$estimate, 1 - x$estimate)
```
etiennebacher, line 20: I think this could be simplified to `out <- ifelse(response != 1, x$estimate, 1 - x$estimate)`

etiennebacher (commented):
Hi @vincentarelbundock, just two minor comments while I was looking for perf improvements

vincentarelbundock:
> I took a quick look (Matt's comments seem right on!) I would just add that it would be useful to be more explicit that your implementation of conformal relies on marginal coverage, as there are other algorithms that provide e.g., class-conditional coverage

vincentarelbundock:
https://arxiv.org/abs/2306.09335

isabellaghement:
Hi Vincent, 

I had a chance to read **https://arelbundock.com/marginaleffects_conformal.html** and have some feedback below.  (I haven't had a chance to play with the actual R code next.) Hopefully, some of it will be useful.  

Your work on adding conformal predictions capabilities to the marginaleffects package is so exciting - thank you on behalf of all of us who do applied work for making things so accessible to us.  

**Naming** 

To me, it seems like _predictions()_ is used to perform inference (e.g., estimation of the expected response value; confidence interval construction for this value) while _inferences()_ does the opposite (e.g., construction of prediction sets).  So the naming of the functions _predictions()_ and _inferences()_ seems backwards?  

There's probably not much that can be done about changing the naming of predictions() at this point.  But it may be possible to change the naming of _inferences()_ to something like _conformalize()_? 

 **Confidence vs. prediction intervals**

I wonder if it may help to provide a visual example here to help outline the difference between these two types of intervals?  Something along these lines: 

image

An example like this would hopefully drive the point that confidence intervals apply to a whole slice of population subjects, whereas prediction intervals apply to a randomly selected individual from the population for whom we know what population slice they belong too.  (The slices in the above example consist of 1. healthy patients; 2. sick patients.)

R/conformal.R
vincentarelbundock: This is kind of tricky because score is not always of the same length as the number of rows (ex: for categorical outcomes, it is one per group). In any case, there are usually less than 25 CV folds, so we are only expanding the vector at most 25 times, which should be negligible. Good observation, though.

R/conformal.R
vincentarelbundock: yep, or `response == 0` to be even more explicit. Thanks!

vincentarelbundock:
Thanks so much @isabellaghement , I really appreciate that you took the time. All good points. Some of them I probably won't be able to address (given time constraints), but I'll do my best.

Cheers!

PR #907, opened 2023-09-20
`estimates()`

vincentarelbundock:


ngreifer:
I beg of you not to do this. You think the maintenance costs are low, but they are not. This is exactly why `Zelig` fell; it tried to do too much. Keep things modular by supporting other packages rather than trying to be a one-stop shop. This will inadvertently require you to be a maintainer of all the packages supported. It is better to provide a vignette explaining how to run each of these models using different packages (which you would likely have to do in a help file for `estimates()`, anyway) than to try to implement them in the package. At best, this should be a separate package that interfaces with `marginaleffects`, not a part of the package. `marginaleffects` is too important of  package to be tethered by this massive weight.

andrewheiss:
I'd agree. I've had to teach {tidymodels} for a couple different workshops and learners have found that it's often too abstract to just change the engine argument. It hides too much of the model-specific interface. Where {tidymodels} shines is in production pipelines where people don't care so much about model debugging and tinkering. When just running a handful of models for a report or manuscript or whatever, the universal {tidymodels} interface is clunky and overkill and hard to work with and its better to just use the regular `glm()`, `brm()`, `ologit()` or whatever functions directly. I'd see similar issues with a {Zelig}-like interface—possibly too abstract

vincentarelbundock:
Very glad I asked you. Thanks both for your input.

I'm convinced and will drop this.

andrewheiss:
It _might_ be useful to go the other way—let {tidymodels} handle all the universal interface heavy lifting and just make sure that `slopes()`, `comparisons()`, and `predictions()` can fit in a pipeline. Julia Silge has an example here of using `predictions()` in a pipeline.

Basically let {tidymodels} be the magical universal interface since they have a whole paid team of developers for that, but ensure that {marginaleffects} can fit in their ecosystem? (I think it already does, since it works with broom)

vincentarelbundock:
Riight. I did write some code to facilitate interaction. In principle, any `tidymodels` object produced by an engine that is supported by `marginaleffects` should also be supported out of the box. Although I haven't tested that extensively.

Issue #893, opened 2023-09-01
`inferences()` should support `hypotheses()` 


Tristan-Siegfried:
I am not sure if I am missing how to do this as it seems like a natural extension of what `marginaleffects` can already do, but I thought it was worth flagging since I didn't see how to do this in the documentation.

The desired usage would be to pass `hypotheses` objects to the `inferences()` function, with the upshot of being able to use e.g. bootstrapping for hypotheses

vincentarelbundock:
That's not currently supported, but it probably should be. If someone is interested in trying to make this work, a good place to start might be to include a call to `inferences_dispatch()` near the very top of the `hypotheses()` function definition. 

Something like this: https://github.com/vincentarelbundock/marginaleffects/blob/main/R/comparisons.R#L406

Somewhere around here: https://github.com/vincentarelbundock/marginaleffects/blob/main/R/hypotheses.R#L134

PR #908, opened 2023-09-21
`inferences()` support for `hypotheses()`

Tristan-Siegfried:
I have taken a crack at the issue discussed in issue #893.

It isn't quite right: It works for bootstrap inferences, but not for simulations. In the simulation based call (not in the initial call), when it tries to calculate
```
se <- get_se_delta(
        model = model,
        vcov = vcov,
        hypothesis = hypothesis,
        FUN = FUNouter,
        numderiv = numderiv,
        ...)
```
The dots it passes contain a list() which breaks `get_se_delta`. I am guessing there is a standard fix for this issue, so I thought it would be more time-efficient to just ask @vincentarelbundock what a reasonable fix for this would be. Note that I tried looking at `comparisons` to see how it avoided this issue compared to `hypotheses`, and on this journey I tried mirroring the syntax when `comparisons` called `get_se_delta`, but the issue persisted for `hypotheses` and so it remains unclear to me why `comparisons` doesn't have this same issue. Would be interested to hear what's going on if you have any ideas.

Also another note: In order for `hypotheses` to play nice with `inferences` and `inferences_dispatch` I needed to rename the `FUN` argument to `hypotheses`. The alternative to changing this one user-facing argument would be to change lots of internal code, so I thought this was a reasonable decision; plus I previously already had the view that `FUN`  was not quite the right name for the argument in `hypotheses`. I have changed that argument to `hypoth_fun`.
Let me know if you see a better way of doing this.


vincentarelbundock:
@Tristan-Siegfried thanks so much for this, it sounds fantastic!

I'm travelling right now, so won't be able to review for 1 or 2 weeks. Sorry for the delay.

From what you describe, it sounds really great. The only thing I'm not (yet) convinced by is the change of argument name. I realize that `FUN` is not pretty, but it is quite standard across the `R` ecosystem. Most importantly, I really want to avoid breaking changes because [some packages](https://github.com/kylebutts) rely on the `FUN` argument. Couldn't we just do something like `hypoth_fun=FUN` at the top of the function to avoid argument conflicts throughout?

vincentarelbundock:
Hmm, I see. `bootstrap_boot` and friends are not user-facing, though. Is it not possible to change the argument name in those internal functions instead?

vincentarelbundock:
Thanks a ton for this! I really appreciate that you took the time to implement it and then went the extra mile to include my suggestions. This looks great and is now merged in `main`. 

I just released a version to CRAN, and they don't want us to submit more than once a month, but the feature will be in the next CRAN release and is now available on Github.

Thanks!

PR #922, opened 2023-10-05
add support for tidymodels workflows, use extractors

simonpcouch:
Apologies for the halfway-there PR, but I figured it might be eliciting nonetheless. :) This implementation speaks to what I believe https://x.com/katsiferis/status/1709971403389198639?s=20 was getting at—I stopped making changes when I realized those needed might be invasive.

parsnip `model_fit`s are one of the fundamental objects in tidymodels, though folks will just as often interface with `workflow`s, which bind `model_fit`s with a preprocessor. I made the minimal changes here to support workflows as well, which would allow folks to hook much more tidymodels output in with marginaleffects. I'll use a formula as a preprocessor for the sake of example.

With those minimal changes, I see:

``` r
library("tidymodels") |> suppressPackageStartupMessages()
data("bikes", package = "fmeffects")
library(marginaleffects)
options(width = 10000)

mod <- workflow(count ~ ., linear_reg()) |>
  fit(data = bikes)
avg_comparisons(mod, newdata = bikes, type = "response")
#> Warning: These variables were not found: seasonspring, seasonsummer, seasonwinter, year1, holidayFalse, weekdayMon, weekdayTue, weekdayWed, weekdayThu, weekdayFri, weekdaySat, workingdayFalse, weathermisty, weatherrain.  Try specifying the `newdata` argument explicitly and make sure the missing variable is included.
#> Warning: Model matrix is rank deficient. Some variance-covariance parameters are missing.
#> Error: Unable to compute predicted values with this model. This error can arise when `insight::get_data()` is unable to extract the dataset from the model object, or when the data frame was modified since fitting the model. You can try to supply a different dataset to the `newdata` argument. If this does not work, you can file a report on the Github Issue Tracker: https://github.com/vincentarelbundock/marginaleffects/issues
```

Understandably, it seems like marginaleffects expects the variables that the underlying (we call it _engine_) model (i.e. `extract_fit_engine(model_fit)` or `extract_fit_engine(workflow)`) had access to will be available in `newdata` when inputted at `predict()` time. When using a workflow, they won't already be there when passed to `predict()`, but the workflow's preprocessor will create them before passing them off to the underlying engine. More concretely, in tidymodels, `fit()` and `predict()` inputs have the same column names—e.g. `"season"`—but when using a preprocessor, the underlying engine fit will only see the pre-processed columns—e.g. `"seasonspring"`, `"seasonsummer"`, `"seasonwinter"`.

If I were implementing, I'd say give workflows a pass here, and they'll do the needed checks and error informatively themselves if need be. If you're not on board for that, one can also extract the pre-processor `pre` and then call `bake(pre, newdata)` to get the data that the parsnip fit (and thus engine fit) expects. Of course, I am likely overlooking some complexity here, and am glad to iterate along with you. :)


R/sanity_model.R
```diff
@@ -131,9 +132,8 @@ sanitize_model <- function(model,
                            vcov = NULL,
                            ...) {
 
-    # tidymodels appear to store the model fit in `model[["fit"]]`
-    if (inherits(model, "model_fit") && "fit" %in% names(model)) {
```
simonpcouch, line 135: parsnip models have class `model_spec` until they are fitted, at which point they become `model_fit`s and always gain that `"fit"` slot.

R/methods_tidymodels.R
```diff
@@ -8,10 +8,21 @@ get_predict.model_fit <- function(model, newdata, type = NULL, ...) {
     return(out)
 }
 
+#' @include get_predict.R
+#' @rdname get_predict
+#' @keywords internal
+#' @export
+get_predict.workflow <- get_predict.model_fit
```
simonpcouch, line 15: `workflow`s and `model_fit`s have type-stable prediction! https://tidymodels.github.io/model-implementation-principles/model-predictions.html#return-values

Seems like this is a TODO already, but predictions will have column names different (but deterministically) than `".pred"` in the classification case. If you just need hard class predictions, that column will be `".pred_class"`. Class probability columns are named w.r.t. the levels of the outcome.

vincentarelbundock:
Thanks for this. It was *super* helpful\!

I still need to work on this, but the last two commits should add `workflow` support and make the `model_fit` support less hack-y (use proper methods rather than object surgery).

This only suppors `numeric`, `prob`, and `class` predictions, but it should be easy to add more by modifying `get_predict.model_fit()`.

``` r
library(tidymodels) |> suppressPackageStartupMessages()
library(marginaleffects)

data("bikes", package = "fmeffects")

# average predictions
# standard errors available because the engine is supported
workflow(count ~ ., linear_reg()) |>
  fit(data = bikes) |>
  predictions(newdata = bikes, type = "numeric", by = "season")
# 
#  season Estimate Std. Error    z Pr(>|z|)     S 2.5 % 97.5 %
#  fall        277       5.13 54.0   <0.001   Inf   267    287
#  spring      115       5.25 22.0   <0.001 352.9   105    126
#  summer      221       5.18 42.8   <0.001   Inf   211    232
#  winter      232       5.30 43.7   <0.001   Inf   221    242
# 
# Columns: season, estimate, std.error, statistic, p.value, s.value, conf.low, conf.high 
# Type:  numeric

# average slopes
# no standard error because the engine is not supported
workflow(count ~ ., rand_forest("regression")) |>
  fit(data = bikes) |>
  avg_slopes(newdata = bikes, type = "numeric")
# 
#        Term      Contrast Estimate
#  count      dY/dX             0.00
#  holiday    False - True     12.72
#  humidity   dY/dX          -209.10
#  month      dY/dX            22.38
#  season     spring - fall   -30.54
#  season     summer - fall    -9.10
#  season     winter - fall     2.02
#  temp       dY/dX            26.59
#  weather    misty - clear    -7.57
#  weather    rain - clear    -59.68
#  weekday    Fri - Sun        72.62
#  weekday    Mon - Sun        81.53
#  weekday    Sat - Sun        23.66
#  weekday    Thu - Sun        88.57
#  weekday    Tue - Sun        86.99
#  weekday    Wed - Sun        89.27
#  windspeed  dY/dX            -4.75
#  workingday False - True   -189.59
#  year       1 - 0            99.44
# 
# Columns: term, contrast, estimate 
# Type:  numeric

# average class probabilities
# TODO: Where are the standard errors?
dat <- transform(mtcars, gear = factor(letters[gear]))
workflow(gear ~ hp * mpg, multinom_reg()) |>
  fit(data = dat) |>
  avg_predictions(newdata = dat, type = "prob")
# 
#  Group Estimate
#      c    0.469
#      d    0.375
#      e    0.156
# 
# Columns: group, estimate 
# Type:  prob

# classes
workflow(gear ~ hp * mpg, multinom_reg()) |>
  fit(data = dat) |>
  predictions(newdata = dat, type = "class")
# 
#  Estimate  mpg cyl  disp  hp drat   wt qsec vs am gear carb
#         d 21.0   6 160.0 110 3.90 2.62 16.5  0  1    d    4
#         d 21.0   6 160.0 110 3.90 2.88 17.0  0  1    d    4
#         d 22.8   4 108.0  93 3.85 2.32 18.6  1  1    d    1
#         d 21.4   6 258.0 110 3.08 3.21 19.4  1  0    c    1
#         c 18.7   8 360.0 175 3.15 3.44 17.0  0  0    c    2
# --- 22 rows omitted. See ?avg_predictions and ?print.marginaleffects --- 
#         e 30.4   4  95.1 113 3.77 1.51 16.9  1  1    e    2
#         e 15.8   8 351.0 264 4.22 3.17 14.5  0  1    e    4
#         e 19.7   6 145.0 175 3.62 2.77 15.5  0  1    e    6
#         e 15.0   8 301.0 335 3.54 3.57 14.6  0  1    e    8
#         d 21.4   4 121.0 109 4.11 2.78 18.6  1  1    d    2
# Columns: rowid, estimate, mpg, cyl, disp, hp, drat, wt, qsec, vs, am, gear, carb 
# Type:  class
```

vincentarelbundock:
Thanks again for this. I really appreciate you taking the time to look at it. It's now merged (with some improvements), and I updated the vignette to show a couple examples with workflows.

Let me know if you find other problems.

I think this is very cool ;)

workflow

simonpcouch:
Wooo! Stoked to see it—thanks for taking the time to consider the PR.

Issue #939, opened 2023-10-20
Feature request: Plotting should allow a fourth variable in `condition` through `facet_grid`


Tristan-Siegfried:
For ease of discussion I will only mention `plot_predictions`, but the comment generalizes to other plotting functions as well.

Currently in `plot_predictions` the `condition` argument takes a list of at most 3 variables on which to condition the predictions, which are mapped as follows: 1: x-axis. 2: color/shape. 3: facets. 
And more specifically the third is mapped to `facet_wrap`.

However, it is possible to provide a nice visualization with a fourth possible variable in `condition` if we visualize the third variable as different columns of a `facet_grid` and visualize the fourth as different rows of that `facet_grid`.

In light of this, I think the following behavior would be an improvement:

1. When 3 variables are passed to `condition` then the current approach is used of passing the third through to `facet_wrap`
2. When 4 variables are passed to `condition` then `facet_grid` is used: the 3rd variable is passed to the `cols` argument of `facet_grid` and the fourth is passed to the `rows` argument of `facet_grid`.

vincentarelbundock:
Yeah, I think that sounds like a nice improvement. I'll be very busy at work in the near future so probably won't have time, but would gladly review a PR if you or anyone else volunteers.

vincentarelbundock:
already merged. thanks again for this.

PR #941, opened 2023-10-20
Add `facet_grid` support with 4 conditions

Tristan-Siegfried:
Add support for 4 conditions and facet_grid, resolving #939 

A nuance here: When labelling facets, if doing `facet_grid` I use `label_both` in order to disambiguate the facet rows/columns, but in the case of `multi_variables` we also want to remove `marginaleffects` internal variable renaming, so do string processing as well.


vincentarelbundock:
Thanks, I'm very excited about this. I'm off the computer now for the weekend, but will look at this next week for sure.

vincentarelbundock:
I had a few minutes this morning and took a look. Overall this looks great!

I updated to `main` and fixed a namespace issue.

There's at least one failing test and I can't really figure out the cause. I think it's this one: https://github.com/vincentarelbundock/marginaleffects/blob/main/inst/tinytest/test-plot_predictions.R#L187

If you want, you can also add your info to `DESCRIPTION` as a `ctb`. (Optional)

nhimwich:
Thanks for this.
Indeed there was a bug, so I'm sorry for that. 
I believe it is now testing all the plotting tests.

vincentarelbundock:
Thanks a lot!

Merged into main now and will be part of the next release.

This is a very nice change.

PR #1014, opened 2024-02-02
Translate equivalence, experiments and gcomputation vignettes to python

LamAdr:
1. **Experiments**
- `cross` comparisons not implemented
- nearly all plots are replicated, not sure if this is needed

2. **Gcomputations**
- I did not translate the boostrap code
- the counterfactual dataset of the Hernán and Robins example contains the original version of the data (interv=-1). If the model is estimated with this semi-counterfactual dataframe, the Python version crashes at `cmp = avg_comparisons(std, variables = {"qsmk" : [0,1]})`. I slightly modified the translation to estimate only on the original data, appending the counterfactual rows after estimation.

3. **Equivalence**
- I did not translate emmeans nor tost
- Giving `predictions()` as an argument to `hypotheses()` seems not implemented?
- `newdata = "median"` causes problems with both `hp` and `gear` if they are not casted with
`mtcars = mtcars.cast({"gear" : pl.Utf8, "hp" : pl.Float64})`.
the problem with `hp` is that it's type is detected as `integer` by datagrid. Note that the problem occurs with
`datagrid(model=mod, FUN_numeric=lambda x: x.median())`
but not with
`datagrid(newdata=mtcars, FUN_numeric=lambda x: x.median())`.
I would have liked to specify that `gear` is categorical in the formula with
`smf.ols("mpg ~ hp + C(gear)", data = mtcars.to_pandas()).fit()`
but this also causes problems with datagrid. I think because it gets its columns from the original data and not from the formula.

Thank you, tell me if you need anything changed


vincentarelbundock:
Thanks a lot that looks fantastic! 

Noted on the issues.

No problem for the things not implemented. Makes sense.

Will check and merge.

I guess you have other things on the todo list already, right?

LamAdr:
> I guess you have other things on the todo list already, right?
Yes, but don't hesitate if something of high priority comes to your mind.

PR #1023, opened 2024-02-14
`glmmTMB`: newparams attempt

vincentarelbundock:
```r
library(marginaleffects)
library(glmmTMB)
pkgload::load_all()
dat <- read.csv("https://vincentarelbundock.github.io/Rdatasets/csv/causaldata/thornton_hiv.csv")
m <- glmmTMB(got ~ age * hiv2004 + (1 | villnum), data = dat)

get_se_manual <- function(m, random = FALSE) {
    s <- TMB::sdreport(m$obj, getJointPrecision = TRUE)
    Sigma <- solve(s$jointPrecision)
    if (isTRUE(random)) {
        XZ <- as.matrix(cbind(getME(m, "X"), getME(m, "Z")))
        keepvars <- c("beta", "b")
    } else {
        XZ <- as.matrix(getME(m, "X"))
        keepvars <- "beta"
    }
    keepvars <- rownames(Sigma) %in% keepvars
    Sigma <- Sigma[keepvars, keepvars]
    se <- sqrt(diag(XZ %*% (Sigma %*% t(XZ))))
    return(se)
}

se1 <- get_se_manual(m, random = FALSE)
se2 <- predict(m, se.fit = TRUE)$se.fit

pkgload::load_all()
p <- predictions(m)
get_vcov(m) |> dim()
get_coef(m) |> length()

b0 <- b1 <- get_coef(m)
b1[2] <- b1[2] + 1e-6
p0 <- predict(m, newparams = b0)
p1 <- predict(m, newparams = b1)
d <- (p1 - p0) / 1e-6
W <- attr(p, "jacobian")
head(d)
head(p0)
head(p1)


attr(p, "jacobian") |> head()
V <- vcov(m)$cond
J <- model.matrix(m)
se3 <- sqrt(diag(J %*% (V %*% t(J))))
se4 <- get_se_manual(m, random = FALSE)

all.equal(unname(se1), unname(se2))
all.equal(unname(se3), unname(se4))

```


DrJerryTAO:
I created these functions that might be useful as building blocks. 

```r
get_Sigma <- function(m, random = FALSE) {
  s <- TMB::sdreport(m$obj, getJointPrecision = TRUE)
  Sigma <- solve(s$jointPrecision)
  if (isTRUE(random)) {
    # XZ <- as.matrix(cbind(getME(m, "X"), getME(m, "Z")))
    keepvars <- c("beta", "b")
  } else {
    # XZ <- as.matrix(getME(m, "X"))
    keepvars <- "beta"
  }
  keepvars <- rownames(Sigma) %in% keepvars
  Sigma <- Sigma[keepvars, keepvars]
  # se <- sqrt(diag(XZ %*% (Sigma %*% t(XZ))))
  return(Sigma)
}

get_XZ <- function(m, random = FALSE) {
  if (isTRUE(random)) {
    XZ <- as.matrix(cbind(getME(m, "X"), getME(m, "Z")))
  } else {
    XZ <- as.matrix(getME(m, "X"))
  }
  return(XZ)
}

get_coef.glmmTMB <- function(model, re.form = NULL, ...) {
  if (is.null(re.form)) {
    b <- model$fit$parfull
  } else {
    b <- model$fit$par
  }
  b <- setNames(as.vector(b), row.names(b))
  return(b)
}

set_coef.glmmTMB <- function(model, coefs, ...) {
  out <- model
  out$fit$parfull <- coefs
  return(out)
}

get_vcov.glmmTMB <- function(model, re.form = NULL, ...) {
  if (is.null(re.form)) {
    sdr <- TMB::sdreport(model$obj, getJointPrecision = TRUE)
    vcov <- solve(sdr$jointPrecision)
    keepvars <- c("beta", "b")
  } else {
    vcov <- Model$sdr$cov.fixed
    keepvars <- "beta"
  }
  keep <- rownames(vcov) %in% keepvars
  vcov <- vcov[keep, keep]
  return(vcov)
}

get_predict.glmmTMB <- function(
    model, newdata = insight::get_data(model), type = "response", ...) {
  # glmmTMB:::predict.glmmTMB()
  out <- stats::predict(model, newdata = newdata, type = type, ...)
  if (is.list(out)) out <- out$fit
  out <- data.frame(rowid = seq_len(nrow(newdata)), estimate = out)
  return(out)
}
```

DrJerryTAO:
With the new functions above, `predictions()` give the correct dimension of `vcov` (including all random effects) as in `View(attributes(predictions(Model, vcov = as.matrix(get_vcov(Model)))))` but misses the Jacobian component. Is there a convenient way to define jacobian for custom model fits?

vincentarelbundock:
No. We need to build with the numeric derivatives.

This exactly the point of my question above.

DrJerryTAO:
The `predict(newparams = )` for numeric-derivatives does not seem promising if the goal is to derive Jacobian considering uncertainty from BLUP. In `glmmTMB:::predict.glmmTMB()`, Line 27 shows the assignment is only to object$fit$par that includes `beta, betad, theta` but not `b` and the new assignment is only used if `fast = FALSE`
```r
  oldPar <- object$fit$par
  if (!is.null(newparams)) 
    oldPar <- newparams
```

Since the problem emerges only in linear mixed models with glmmTMB, the derivative of the response to each coefficient (fixed effects plus BLUP random effective, not including log standard deviation of random effects or log variance of residuals) should be simply the design matrix, right? If so, is the problem actually about extracting the design matrix with random-effect predictors? `model.matrix` will not do this correctly, as it only extracts fixed predictors according to `glmmTMB:::model.matrix.glmmTMB(): Error only fixed model matrices currently available`. I can find no other functions to extract the design matrix in glmmTMB. I found the best tool to extract (and generate new) fixed and random design matrix (called `XZ` in your `get_se_manual `) is
```r
cbind(
  lme4::glFormula(
    Model$modelInfo$allForm$formula, data = model.frame(Model), 
    contrasts = Model$modelInfo$contrasts)$X, 
  lme4::glFormula(
    Model$modelInfo$allForm$formula, data = model.frame(Model), 
    contrasts = Model$modelInfo$contrasts)$reTrms$Zt |> t() |> as.matrix())
```
This only deals with the conditional mean component, as formulae for conditional `$allForm$formula`, zero-inflation `$allForm$ziformula`, and dispersion `$allForm$dispformula` are stored in different list element. The combined version `$allForm$combForm` is not usable as it removes `|` in the random term, resulting in `got ~ age * hiv2004 + (1 + villnum) + 0 + 1`. Since they are all generated from the same data frame, this method guarantees that the same predictor with both fixed and random effects are changed all together, if the data manipulation is done to `model.frame(Model)`. Nevertheless, I think we can skip the zero-inflation model for linear mixed models. The dispersion model in linear mixed models is for heteroscedasticity of residuals, so I do not think it needs taking derives of.

vincentarelbundock:
The design matrix is just one special case. Things are not so simple for slopes and contrasts.

vincentarelbundock:
```r
get_predict.default <- function(model,
                                newdata = insight::get_data(model),
                                type = "response",
                                newparams = NULL,
                                ...) {

    dots <- list(...)
@@ -32,6 +33,7 @@ get_predict.default <- function(model,
    # first argument in the predict methods is not always named "x" or "model"
    dots[["newdata"]] <- newdata
    dots[["type"]] <- type
    dots[["newparams"]] <- newparams
    args <- c(list(model), dots)

    # `pred` is a secret argument ca
```


In `get_se_delta.R`:

```r
       # glmmTMB needs this
        args$newparams <- x

        g <- do.call("FUN", args)
```

```r

get_predict.glmmTMB <- function(model,
                                newdata = insight::get_data(model),
                                type = "response",
                                newparams = NULL,
                                ...) {

    if (inherits(vcov, "vcov.glmmTMB")) {
@@ -16,6 +17,7 @@ get_predict.glmmTMB <- function(model,
        newdata = newdata,
        type = type,
        allow.new.levels = TRUE, # otherwise we get errors in marginal_means()
        newparams = newparams,
        ...)
    return(out)
}
```

vincentarelbundock:
I tried again to make this PR work but failed. I am not giving up. 

I'd be happy to merge a pull request and to provide support in development. if an expert can solve this problem. however, I will no longer work on this myself.

Issue #1046, opened 2024-03-13
Feature request: automatic extraction of weights from fitted model


ngreifer:
Back with another suggestion related to weights. When we fit a model with weights and want to use `avg_predictions()`, etc., to compute the weighted means, we need to supply the name of the weights variable in the dataset to `wts` or a vector containing the weights. But the weights are already in the model; I'm thinking it might be easy to allow users to set `wts = TRUE` to automatically extract the weights in the model using, e.g., `insight::get_weightit()` or `stats::weights()`. I think the default should be what it is now (no weights used even if included in the model fit), but I think it is worth having it as an option. In Stata, `margins` automatically applies weights used to fit the model when computing averages; the user has to turn this off by setting `, noweights`. I think it's okay to have the default here be different from Stata, but I think it would also be beneficial to have a simple argument that automatically extracts the weights already present within the object rather than requiring the user to have to specify the name of the weights variable. This would be especially useful when the weights variable is not in the supplied dataset, but is in the model, e.g., with `svyglm()`, which appends a `"(weights)"` variable to the model frame, typically invisible to the user.

It might be that setting `wts = TRUE` only works when `newdata` is not supplied. This goes kind of in combination with #1045.

For example, I'm imagining a user could call
```r
avg_comparisons(fit, variables = "treat", wts = TRUE, subset = treat == 1)
```
to compute a weighted ATT without ever having to supply the dataset again or known the name of the weights column in the dataset (if it's even there).

vincentarelbundock:
I like this a lot, but might not have the bandwidth to implement and test very soon.

For later reference, I'm optimistic that a slight tweak to this function (and documentation) would do the trick:

https://github.com/vincentarelbundock/marginaleffects/blob/51f8fe6def297dce5ac772880b709da4f23cdf20/R/sanitize_newdata.R#L93

ngreifer:
I agree that this could be easy, but I'll let you be the judge of that :) I think adding
```r
if (isTRUE(wts)) wts <- insight::get_weights(model)
```
in `sanitize_newdata()` would do it.

vincentarelbundock:
https://github.com/vincentarelbundock/marginaleffects/pull/1050

PR #1049, opened 2024-03-15
Allowing `wts = TRUE` to extract weights from model using `insight::get_weights()`

ngreifer:
See #1046. There are still some issues to work out:
* When `newdata` is supplied as something other than the original model data, the extracted weights will not have the same length as the supplied `newdata` and will cause an error.
* When `newdata` is supplied as a call to `datagrid()`, the weights are appended after evaluation, when `modeldata` should have the weights appended before being evaluated by `datagrid()`. This leads to the error above.
* The error above is not clear; the user will see that the length of `wts` must match that of `newdata`.
* I didn't re-build the documentation.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: marginaleffects
 Title: Predictions, Comparisons, Slopes, Marginal Means, and Hypothesis Tests
-Version: 0.18.0.9003
+Version: 0.18.0.9004
 Authors@R: 
     c(person(given = "Vincent",
              family = "Arel-Bundock",
--- a/R/sanitize_newdata.R
+++ b/R/sanitize_newdata.R
@@ -103,7 +103,7 @@ add_wts_column <- function(wts, newdata) {
             stop(msg, call. = FALSE)
         }
     }
-    
+
     # weights: before sanitize_variables
     if (!is.null(wts) && isTRUE(checkmate::check_string(wts))) {
         newdata[["marginaleffects_wts_internal"]] <- newdata[[wts]]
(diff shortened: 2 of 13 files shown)


vincentarelbundock:
> * When `newdata` is supplied as something other than the original model data, the extracted weights will not have the same length as the supplied `newdata` and will cause an error.

Can we not insert a length assertion with an informative error message there?

> * When `newdata` is supplied as a call to `datagrid()`, the weights are appended after evaluation, when `modeldata` should have the weights appended before being evaluated by `datagrid()`. This leads to the error above.

What is a use-case where we want weights to be appended before evaluating `datagrid()`? Do we ever want to use the mean or mode of weights, which is what `datagrid()` will do?

> * I didn't re-build the documentation.

I pushed new docs.

ngreifer:
You know `datagrid(., grid_type = "counterfactual")` is the only `datagrid()` for me :) I mean that let's say I want something like
```r
avg_predictions(fit, newdata = datagrid(x1 = 0:1, x2 = 0:1, grid_type = "counterfactual"),
                by = c("x1", "x2"), wts = TRUE)
```
Because the `newdata` is 4 times the length of the model dataset, the weights extracted from the model don't have the same length as the data. But since the model data is just being replicated 4 times, the weights should be appended to the model data first. That way they remain the same length as the dataset.

For other uses of `datagrid()`, this doesn't matter. It's only when `datagrid` is using the model data that this matters. For your eventual implementation of subsetting (🙏🏻) it will also be useful to append the weights to the model data so that the weights are automatically subsetted.

vincentarelbundock:
A more promising strategy might be to do this earlier in the process: in `get_modeldata()`. 

I'm doing family stuff so won't be able to look at this for a while, but here's a pull request in case you want to tinker: https://github.com/vincentarelbundock/marginaleffects/pull/1050

vincentarelbundock:
Thanks again for the attempt. I really appreciate it.

Should be solved by: https://github.com/vincentarelbundock/marginaleffects/pull/1050

Don't hesitate to open a new issue if you see problems.

PR #1121, opened 2024-05-06
Issue #1120: automatically shift to `ratioavg` when calling `avg_comparisons()`

vincentarelbundock:
@ngreifer, do you have a view of what the default ratio function should be when we take an average?

``` r
library(marginaleffects)
mod <- glm(vs ~ am + wt, data = mtcars, family = binomial)
d0 <- transform(mtcars, am = 0)
d1 <- transform(mtcars, am = 1)
p0 <- predict(mod, newdata = d0, type = "response")
p1 <- predict(mod, newdata = d1, type = "response")

# marginaleffects 0.20.0
mean(p1 / p0) 

    [1] 0.2375687

# this PR
mean(p1) / mean(p0) 

    [1] 0.395687

avg_comparisons(mod, variables = "am", comparison = "ratio")


     Term          Contrast Estimate Std. Error    z Pr(>|z|)    S 2.5 % 97.5 %
       am mean(1) / mean(0)    0.396     0.0779 5.08   <0.001 21.3 0.243  0.548
```


ngreifer:
It should absolutely be the latter, but I thought users could already request this using `"ratioavg"`. The former is not useful because risk ratios are not collapsible (except in a very obscure sense). I think it would be a good change to make all the `avg_{.}()` functions automatically apply the `"{.}avg"` suffix for any comparison.

PR #1171, opened 2024-07-15
Fix some lints

etiennebacher:
Hi @vincentarelbundock, I've made a new package to find lints and automatically fix them ([`flint`](https://flint.etiennebacher.com/)). I used `marginaleffects` as one of my test cases and found a few lints so I thought you might want to remove them. In particular, I think it might have fixed a bug in `R/methods_nnet.R`.

FYI those fixes were automatically generated with `flint::fix_dir("R", exclude_linters = "equal_assignment")`. End of the self-advertising :)

### DIFF
--- a/R/complete_levels.R
+++ b/R/complete_levels.R
@@ -33,7 +33,7 @@ complete_levels <- function(x, character_levels = NULL) {
         # HACK: Some models use a character variable with many levels (e.g.,
         # mixed-effects groups). This creates a massive padding dataset, and making
         # predictions can become very expensive.
-        if (isTRUE(sum(sapply(vault, length)) > 100)) return(data.frame())
+        if (isTRUE(sum(lengths(vault)) > 100)) return(data.frame())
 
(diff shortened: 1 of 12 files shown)


vincentarelbundock:
Thanks @etiennebacher. The new tool looks great.

Yes, this is definitely a bug. Good catch.

Unfortunately, it also looks like `flint` introduced a new bug—or at least a departure from intended behavior.

``` r
model <- glm(mpg ~ hp, data = mtcars)

# same behavior
sup <- c("glm", "lm")
isTRUE(all(sup %in% class(model)))
> [1] TRUE
isTRUE(all(inherits(model, sup)))
> [1] TRUE

# different behavior
sup <- c("glm", "junk")
isTRUE(all(sup %in% class(model)))
> [1] FALSE
isTRUE(all(inherits(model, sup)))
> [1] TRUE
```

etiennebacher:
Thanks, nice catch, I reverted the change here and added a fix in `flint`

PR #1177, opened 2024-07-19
Fix more lints

etiennebacher:
Hi @vincentarelbundock, I've added [more linting rules](https://github.com/etiennebacher/flint/releases/tag/v0.0.2) in `flint` and here are a few additional fixes. Most of them are replacing `any(!expr)` by `!all(expr)` and adding `0` in front of the decimal when it's missing.

I'm testing on this package so for me there's no cost in proposing those changes, but of course let me know if you'd prefer not having this kind of PR.

### DIFF
--- a/R/ci.R
+++ b/R/ci.R
@@ -21,7 +21,7 @@ get_ci <- function(
     }
 
     required <- c("estimate", "std.error")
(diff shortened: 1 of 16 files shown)


vincentarelbundock:
These all look like improvements. Love those types of PRs!

Thanks!

PR #1182, opened 2024-07-26
Added support for new WeightIt models

ngreifer:
`WeightIt` has a few new model classes for outcome models that incorporate uncertainty in the estimation of the weights, if included. The functions are `glm_weightit()`, which is a wrapper for `glm()`, `coxph_weightit()`, which is a wrapper for `coxph()`, `multinom_weightit()`, and `ordinal_weightit()`. These latter two have their own predict methods. I updated the type dictionary, added the models to supported classes, and white-listed the `values` argument to the `predict()` methods for these models. Extracting coefficients, covariance matrices, and predictions using `coef()`, `vcov()`, and `predict()`, respectively, work without modification using the default methods.


vincentarelbundock:
Sounds great. 

Can you add a note in NEWS?

After that, I'll happily merge when there are no merge conflicts.

PR #1229, opened 2024-10-09
WIP: add benchmark workflow

etiennebacher:
Starting setting up the benchmark workflow using `atime` for now, but using `cross` + `bench` is also an option.

Not ready at all for now, I'm just trying to get something working and it's easier with an open PR, and then we can discuss about what kind of benchmarks should be included.


etiennebacher:
Seems that `atime` does some weird things related to environments and manipulating expressions. I'll move on to use `cross` and `bench` (I'm also not the biggest fan of `atime` visualizations).

vincentarelbundock:
Sounds great, thanks!

etiennebacher:
@vincentarelbundock the main setup is done but I need to have this merged before finishing the github workflow to post the benchmark results as a comment in each PR. The main issue here is that posting a comment in a PR requires the permission to write, even for PR that are created from forks which makes it more complicated.

This is quite nicely explained in this blog post: https://jacobtomlinson.dev/posts/2022/commenting-on-pull-requests-with-github-actions/

Hopefully the next PR will be sufficient to get the workflow running.

PR #1246, opened 2024-10-23
CI benchmark: produce table instead of plot

etiennebacher:
Gives this (with only 10 and 50 obs):

image

### DIFF
--- a/.github/scripts/benchmarks.R
+++ b/.github/scripts/benchmarks.R
@@ -16,7 +16,7 @@ out <- cross::run(
     library(data.table)
 
     bench::press(
-      N = c(10, 50),
+      N = c(100, 500, 1000, 10000, 50000),
       {
         dat <- data.frame(matrix(rnorm(N * 26), ncol = 26))
         mod <- lm(X1 ~ ., dat)
@@ -32,7 +32,7 @@ out <- cross::run(
           # 26 variables; no standard error
           slopes(mod, vcov = FALSE),
           # 26 variables
-          marginaleffects::slopes(mod),
+          slopes(mod),
           check = FALSE,
           iterations = 5
         )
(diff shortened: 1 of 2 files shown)


etiennebacher:
And now we also have the confirmation that the CI works for PR from forks as well

etiennebacher:
Images are only useful if we have a few benchmarks (6 looks like a nice limit to me). If we want more, then producing a table makes more sense to me, something like this: https://github.com/pola-rs/polars/pull/18473#issuecomment-2320965250

The main question now is do you have specific things that you want to benchmark? I took those "slopes" benchmarks from the "Performance" chapter, where there are a couple of benchmarks run on "predictions" as well. In previous PRs about performance I only used examples that are displayed throughout the docs, I could reuse them for benchmarks.

Overall we won't be able to test all arguments, but I think having at least 1/2 benchmark for each of the main functions makes sense, at least to ensure there's no "catastrophic" performance downgrade.

> Are the facets "free" for commands that take much more time.

Yes the y-scale varies depending on the min/max for each benchmark. It's just not visible in the graphs above because the initial time is very close to 0 for all benchmarks so the scales are the same.

vincentarelbundock:
Yeah, that makes sense. I think a table is probably better, cause we'll probably want ~40 benchmarks at least to cover a good chunk of functionality.

I think the core commands we want to test are those from these chapters:

- https://marginaleffects.com/chapters/hypothesis.html
- https://marginaleffects.com/chapters/predictions.html
- https://marginaleffects.com/chapters/comparisons.html
- https://marginaleffects.com/chapters/slopes.html

Just need to make sure the data used are big enough, otherwise we'll just be comparing overhead, and that will be constant.

vincentarelbundock:
PS: `tinytable` can now export to Github-flavored markdown, so that might be an easy way to produce the table.

etiennebacher:
Ok, I'll update this PR to produce a table and I'll add some benchmarks

etiennebacher:
I'd need this to be merged so that I can check in another PR that the table is well rendered (it's the drawback of this CI workflow, it always runs the version on `main` so any change to the workflow needs to be merged before it can applied, even in the PR that applies those changes).

etiennebacher:
Nevermind, I'm experimenting on a separate repo

etiennebacher:
Can you take a look at https://github.com/etiennebacher/my_marginaleffects/pull/5 and let me know what you think? Differences in timing will be more precise as the number of iterations (10 for now) and the size of the data (75k obs for now) increase, but that will slow down the workflow. For now it runs in 11min.

vincentarelbundock:
This looks amazing! Love the little explosions. That's a nice touch.

I don't mind the run time so much. It's just a reality, and there's nothing we can do.

etiennebacher:
I have updated the PR I linked above to also compare the results with CRAN. This is because many commits are pushed to main directly, which might hide performance issues, for example if we have:

* commit that slows down the code pushed to main directly
* commit that doesn't affect the code is pushed to a PR, therefore running benchmark
* benchmark doesn't show any difference between PR and main, but the overall performance decreased compared to CRAN because of the commit pushed directly to main

The benchmark now has 75k obs and 20 iterations, and runs in 25 min. Additional question: should the CI fail if there is at least one performance decrease? I would say no because there can be spurious differences, but maybe you would like to be notified if this is the case? If not and if you're happy with the output in the linked PR above, then you can merge this.

PR #1250, opened 2024-10-27
CI benchmark: add docs

### LINKED ISSUE #1248: Benchmarking docs
@etiennebacher 

The new benchmarking action is *super* cool. I am very, very pleased with this new addition. Thanks so much!

I think the only two things I'd be curious about are:

1. How to trigger the action again on an existing PR (see my attempt https://github.com/vincentarelbundock/marginaleffects/pull/1244)
2. Would it be possible to "document" or to describe the workflow in a just a few bullet points. My guess is that people would be interested in a detailed tutorial, but if we at least have some bullet points, I could take care of writing the long form version eventually.

etiennebacher:
(and some automatic code styling)

Do you want something a bit more detailed?

Close #1248

### DIFF
--- a/.github/benchmark_docs.md
+++ b/.github/benchmark_docs.md
@@ -1,0 +1,70 @@
+Every time a PR is opened and a commit is added to the PR, a Github Actions 
+workflow is run with a suite of benchmarks. This document explains how this
+workflow works.
+
+# Defining the workflow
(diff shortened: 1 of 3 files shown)


vincentarelbundock:
This is perfect, thanks!

PR #1298, opened 2024-12-08
Some fixes for lints

etiennebacher:
Just the occasional run of `flint` to check recent changes (both in `flint` and in `marginaleffects`).


vincentarelbundock:
Thanks!

PR #1309, opened 2024-12-26
`hypotheses()`: no special treatment for internal objects

vincentarelbundock:
This might be a breaking change. But having no special treatment for internal objects allows us to pass a different `unconditional` to the `vcov` argument.

`recall()` is also dangerous, and I never liked the design. Best to induce people to use the `hypothesis` argument instead.


vincentarelbundock:
Doesn't work:

```r
hypotheses(p, hypothesis = ~reference, vcov = V)
```

PR #1334, opened 2025-01-17
Better concatenation of warning message

MichaelChirico:
Observed in data.table's revdeps checker logs:

```
It is safer and faster to convert such variables to factor before fitting the model and calling a `marginaleffects` function.This warning appears once per session.
```

Note that `warning()` concatenates its inputs with `''`:

```r
warning(letters)
# Warning message:
# abcdefghijklmnopqrstuvwxyz 
```

I'm not familiar with how `insight::format_warning()` handles vectors, so I opted to just build the string directly.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -56,14 +56,13 @@ set_marginaleffects_attributes <- function(x, attr_cache, prefix = "") {
 
 
 warn_once <- function(msg, id) {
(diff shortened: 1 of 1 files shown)


vincentarelbundock:
Thanks @MichaelChirico . I really appreciate that you took the time to write this PR!

PR #1362, opened 2025-01-25
hypothesis -> term column

vincentarelbundock:


vincentarelbundock:
we do need both the `hypothesis` and `term` columns.

PR #1377, opened 2025-01-31
Attempt to fix #1117

### LINKED ISSUE #1117: `newdata = subset()` with multiple imputation
Trying to estimate the ATT after multiple imputation using g-computation:

``` r
library(marginaleffects); library(mice)
data("lalonde_mis", package = "cobalt")

imp <- mice(lalonde_mis, print = FALSE)

est <- with(imp, lm(re78 ~ treat * (age + educ + re74)))

avg_predictions(est, variables = "treat",
                newdata = subset(treat == 1))
#> Error: object 'treat' not found
```

<sup>Created on 2024-05-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

I think ideally, the `subset()` functionality would apply to each imputed dataset, which should be extractable from the `mira` object. This is the source of the bug observed [here](https://stats.stackexchange.com/q/646546/116195).

ngreifer:
Hi Vincent,

I made some changes to attempt to fix #1117. Below are some notes on the changes I made.

* Wrote a new function, `construct_call()` in construct_call.R, which is similar to `match.call()` but works well in `lapply()`, etc. Its output is a call with all arguments unevaluated except the model object, which means to use the arguments you need to evaluate them unless you evaluate the whole call. This makes it possible to store unevaluated expressions like in `newdata`, e.g., `newdata = subset(treat == 1)`, which can be passed to other functions before evaluating it. The output is passed to the `call` component of the output objects and is also used in `process_imputation()` to call the estimation function on each model. `construct_call()` requires an argument identifying the calling function as a string; that function is searched for in the package namespace and the names of its arguments extracted. Then the values of each argument are extracted from the original function call using `substitute()`, with default values left as they are. Values in `...` are also appended.

* A consequence of the above is that we should be using call objects to call other functions when supplying the same arguments rather than calling them as if the arguments had been evaluated. For example, in `avg_predictions()`, we call `construct_call(model, "predictions")` immediately and then evaluate the call rather than passing arguments from `avg_predictions()` to `predictions()`. This means we don't have to perform any processing in `avg_predictions()`. that can occur in `predictions()`.

* For `slopes()`, which passes its input to `comparisons()`, we don't want to pass `newdata` to `sanitize_model()` because it hasn't been properly evaluated yet, since that happens in `comparisons()`. Fortunately, `newdata` isn't used in any of the `sanitize_model_specific()` methods, so I removed `newdata` from being passed to it in `slopes()`.

* It can be expensive to evaluate `...`, so I used new R functions `...names()`, `...elt()`, and `...length()` here and there to reduce the need for `list(...)`. In particular, `...elt(match("effect", ...names())[1L])` is equivalent to `list(...)[["effect"]]`, `"effect" %in% ...names()` is equivalent to `"effect" %in% names(list(...))`, and `...length()` is equivalent to `length(list(...))`.

* I changed `args <- c(args, dots)` to the more modern `args <- utils::modifyList(args, dots)`, which was used in some places already. I left `args <- c(dots, args)` alone in case you had a good reason for it.

I don't know how to use `tinytest` so I wasn't able to test whether anything was broken, but I ran the functions, including in `lapply()`, to make sure things were as expected, and they were. My main worry is interactions with `inferences()`.

I didn't add a NEWS item or any tests because I didn't know if you'll accept this since it's somewhat structural, but at least it's an example of how the issues could be fixed.

Noah


R/imputation.R
```diff
@@ -40,20 +49,26 @@ process_imputation <- function(x, call_attr, marginal_means = FALSE) {
     out$df <- mipool$pooled$df
   }
   out$std.error <- sqrt(mipool$pooled$t)
+
   out <- get_ci(
     out,
-    vcov = call_attr[["vcov"]],
+    vcov = NULL,
```
vincentarelbundock, line 55: aren't we losing information here? Maybe be unrelated, but there's an error in `test-inferences.R` about missing `vcov`

vincentarelbundock:
Wow, thanks, this sounds super interesting. It's pretty deep stuff, and life is crazy right now, so it might take me a bit to wrap my brain around this PR.

Meanwhile, run all tests from the command line:

```sh
make testall
```

Run all tests from R (root of pkg working directory):

```r
pkgload::load_all()
tinytest::run_test_dir()
```

Run one test file from R:

```r
pkgload::load_all()
tinytest::run_test_file("inst/tinytest/test-inferences.R")
```

vincentarelbundock:
I updated a few tests. There is definitely something broken with the `vcov` argument in one of the `test-inferences.R` tests.

There also seems to be a scoping issue with the `expect_slopes()` function, especially when called in the `test-pkg-AER.R` and `test-pkg-brglm2.R` tests.

R/imputation.R
ngreifer: I think it's unrelated because in `get_ci()`, `vcov` is only used to determine whether to compute the z-statistic and p-value, but in this context, other conditions guarantee they are computed, so `vcov` isn't used at all.

vincentarelbundock: Yeah, that seems right. There's still a vcov issue in test-inferences.R, but it's almost certainly a different source.

ngreifer:
I think a possible issue is that we need to better distinguish between the components of the unevaluated call object produced by `construct_call()` and the output components computed and stored in the output object. But there may be some occasions where we just want to evaluate one component of the call object at a time rather than the whole thing. As I understand it, previously you evaluated all the arguments and then inserted those into the call object to avoid using `match.call()`. This means you had to evaluate all components of the call before constructing the call object. A benefit of this is you could use components of the call object as computed values in downstream functions. I made some attempt to remove that when not necessary, like the `vcov` issue in `get_ci()` when called from `process_imputation()`, but clearly I couldn't make those changes everywhere necessary.

For `inferences()`, I think it becomes important to distinguish the computed values stored as attributes of the output object from the unevaluated components of the call object, and I would bet these issues would be solved doing that. I can look into it myself as well.

The code in `construct_call()` is a bit obscure, so I'll try to explain it. Ideally, what I want is to be able to run
```
fun <- function(x) {
  substitute(x)
}
fun(dont_eval_me(bad))
```
which should return a language object `dont_eval_me(bad)` which is not evaluated (so it doesn't matter if it is syntactic). I want to do that for every argument, but unfortunately it isn't easy to loop through argument names and simply run `substitute()`. Instead, I create a string using `sprintf()`, e.g., `sprintf("substitute(%s)", i)`, turn that into a language object, and then evaluate it. Because I'm doing that in an auxiliary function, I need to provide the correct environment to do this, which is in the parent environment.

This approach manually constructs the call, similar to your existing approach, but does so without requiring you to manually program the argument names (because it loops through all of them for the supplied function) and without evaluating the arguments. Because it is manually constructed, it works well inside `lapply()`.

This also allows for passing more unevaluated arguments beyond just `newdata` should you want to do that in the future. For example, if you add a `subset` argument, it also would not be evaluated until you wanted it to be, so you could use semantics similar to what `lm()` uses to evaluate its `subset` argument (i.e., it is evaluated in the context of the dataset). Similarly, you could make it easier to do things like `vcov = vcovCL(cluster = ~school, type = "HC0")` for `vcov`, where the model object is automatically supplied to the first argument, rather than requiring an anonymous function like `vcov = \(x) vcovCL(cluster = ~school, type = "HC0")`. Just some thoughts :)

R/utils.R
```diff
@@ -155,3 +157,8 @@ group_to_factor <- function(group, model) {
     }
     return(group)
 }
+
+
+...get <- function(name, ...) {
+    ...elt(match(name, ...names())[1L])
+}
```
ngreifer, line 164: I see you added this function, which I have been using in some of my packages. I now use a different version:
```
...get <- function(x, ifnotfound = NULL) {
  eval(quote(if (!anyNA(.m1 <- match(.x, ...names())) && is_not_null(.m2 <- ...elt(.m1))) .m2
             else .ifnotfound),
       pairlist(.x = x[1L], .ifnotfound = ifnotfound),
       parent.frame(1L))
}
```
This allows you not to have to pass `...` every time, making it more like `get()` and `...elt()`. I would guess R will eventually add a function like this as a default function. It is unfortunately a bit slower than the version you are using, but the difference is small enough that I would prefer to use the simpler syntax omitting `...`.

I also made a function `...mget()` for getting a list, since `...get()` doesn't work well in `lapply()`.
```
...mget <- function(x) {
  found <- match(x, eval(quote(...names()), parent.frame(1L)))

  not_found <- is.na(found)

  if (all(not_found)) {
    return(list())
  }

  setNames(lapply(found[!not_found], function(z) {
    eval(quote(...elt(.z)),
         pairlist(.z = z),
         parent.frame(3L))
  }), x[!not_found])
}
```
Feel free to use or edit either without any need for attribution.

vincentarelbundock: looks nice. The problem with these is that they were introduced in 4.1, so we need to onboard `backports` as a dependency.

vincentarelbundock:
That’s interesting and useful. It opens up interesting possibilities.

I’ll admit that I’m quite uncomfortable with the `sprintf()` trick, but I imagine you’ve looked at alternatives.

In the short run, here are the two main problems are see.

``` r
pkgload::load_all()
#> ℹ Loading marginaleffects
set.seed(1234)
R <- 1000
mod <- lm(Petal.Length ~ Sepal.Length * Sepal.Width, data = iris)
```

### Bootstrap uncertainty is now bigger than delta method. This was not the case before.

``` r
slopes(mod) |>
        inferences(method = "boot", R = R) |>
        head(3)
#> 
#>  Estimate Std. Error 2.5 % 97.5 %
#>      1.97      0.251  1.31   2.31
#>      1.72      0.259  1.23   2.32
#>      1.82      0.266  1.26   2.40
#> 
#> Term: Sepal.Length
#> Type:  response 
#> Comparison: dY/dX

slopes(mod) |> head(3)
#> 
#>  Estimate Std. Error    z Pr(>|z|)     S 2.5 % 97.5 %
#>      1.97     0.0918 21.5   <0.001 338.5  1.79   2.15
#>      1.72     0.0653 26.4   <0.001 506.4  1.59   1.85
#>      1.82     0.0647 28.2   <0.001 577.8  1.70   1.95
#> 
#> Term: Sepal.Length
#> Type:  response 
#> Comparison: dY/dX
```

### Error (not just a test failure):

``` r
mod <- lm(mpg ~ hp + cyl, data = mtcars)
hypotheses(mod, hypothesis = "hp/cyl=1") |> inferences(method = "simulation", R = 25)
#> Error in (function (model, INF_FUN, vcov, ...) : argument "vcov" is missing, with no default
```

vincentarelbundock:
@ngreifer FYI, it looks like we can nest `substitute()` calls to avoid using `sprintf()`

```r
construct_call <- function(model, calling_function, env = parent.frame(1L)) {
    call_attr <- list(
        name = calling_function,
        model = model
    )

    all_arg_names <- names(formals(get(calling_function), asNamespace(utils::packageName())))
    arg_names <- setdiff(all_arg_names, c("model", "..."))

    call_attr <- c(
        call_attr,
        stats::setNames(
            lapply(arg_names, function(i) {
                eval(substitute(substitute(arg), list(arg = as.name(i))), envir = env)
            }),
            arg_names
        )
    )

    if ("..." %in% all_arg_names) {
        call_attr <- c(call_attr, as.list(substitute(alist(...), env))[-1L])
    }

    do.call("call", call_attr, quote = TRUE)
}
```

vincentarelbundock:
merge

thanks again. this is a nice improvement

ngreifer:
> @ngreifer FYI, it looks like we can nest `substitute()` calls to avoid using `sprintf()`
> 
> ```r
> construct_call <- function(model, calling_function, env = parent.frame(1L)) {
>     call_attr <- list(
>         name = calling_function,
>         model = model
>     )
> 
>     all_arg_names <- names(formals(get(calling_function), asNamespace(utils::packageName())))
>     arg_names <- setdiff(all_arg_names, c("model", "..."))
> 
>     call_attr <- c(
>         call_attr,
>         stats::setNames(
>             lapply(arg_names, function(i) {
>                 eval(substitute(substitute(arg), list(arg = as.name(i))), envir = env)
>             }),
>             arg_names
>         )
>     )
> 
>     if ("..." %in% all_arg_names) {
>         call_attr <- c(call_attr, as.list(substitute(alist(...), env))[-1L])
>     }
> 
>     do.call("call", call_attr, quote = TRUE)
> }
> ```

Wow, among the many many combinations I attempted, I did not find this one. Glad it worked!

vincentarelbundock:
Transparency: it was the first GPT suggestion

ngreifer:
I actually asked Chat GPT about this too but couldn't get it to work for whatever reason. I guess I messed up somewhere.

vincentarelbundock:
[surprised]     Vincent Arel-Bundock reacted to your message:
________________________________

Issue #1388, opened 2025-02-14
`invlink(link)` option for glmmTMB models?


strengejacke:
For glmmTMB models, the "response" calculates "inaccurate" confidence intervals outside the possible/plausible bounds, see example below. Would it be possible to have an `"invlink(link)"` option like for usual glm's also for glmmTMB?

``` r
data <- mtcars
data$count <- rep(c(0, 0, 0, 0, 1, 2, 4), length.out=nrow(mtcars))

# GLM case

m <- glm(count ~ cyl, data=datawizard::data_modify(data, cyl = as.factor(cyl)), family="poisson")
marginaleffects::avg_predictions(m, newdata = "balanced", by = "cyl", type = "invlink(link)")
#> 
#>  cyl Estimate Pr(>|z|)   S  2.5 % 97.5 %
#>    4    1.273   0.3669 1.4 0.7538   2.15
#>    6    0.286   0.0764 3.7 0.0715   1.14
#>    8    0.857   0.5933 0.8 0.4868   1.51
#> 
#> Type:  invlink(link)

emmeans::emmeans(m, "cyl", type = "response")
#>  cyl  rate    SE  df asymp.LCL asymp.UCL
#>  4   1.273 0.340 Inf    0.7538      2.15
#>  6   0.286 0.202 Inf    0.0715      1.14
#>  8   0.857 0.247 Inf    0.4868      1.51
#> 
#> Confidence level used: 0.95 
#> Intervals are back-transformed from the log scale

# glmmTMB case

m <- glmmTMB::glmmTMB(count ~ cyl, data=datawizard::data_modify(data, cyl = as.factor(cyl)), family="poisson")
marginaleffects::avg_predictions(m, newdata = "balanced", by = "cyl", type = "invlink(link)")
#> Error in sanitize_type(model = model, type = type, by = by, calling_function = "predictions"): Assertion on 'type' failed: Must be element of set {'response','link','conditional','zprob','zlink','disp'}, but is 'invlink(link)'.

marginaleffects::avg_predictions(m, newdata = "balanced", by = "cyl")
#> 
#>  cyl Estimate Std. Error    z Pr(>|z|)    S  2.5 % 97.5 %
#>    4    1.273      0.340 3.74   <0.001 12.4  0.606  1.939
#>    6    0.286      0.202 1.41    0.157  2.7 -0.110  0.682
#>    8    0.857      0.247 3.46   <0.001 10.9  0.372  1.342
#> 
[...]

PR #1395, opened 2025-02-16
`invlink(link)` option for glmmTMB models

strengejacke:
Fixes https://github.com/vincentarelbundock/marginaleffects/issues/1388

### DIFF
--- a/R/sanitize_type.R
+++ b/R/sanitize_type.R
@@ -13,7 +13,7 @@ sanitize_type <- function(model, type, by = FALSE, calling_function = "raw") {
     }
 
     checkmate::assert_character(type, len = 1, null.ok = TRUE)
-    
+
     if (inherits(model, "model_fit")) {
         cl <- "model_fit"
     } else {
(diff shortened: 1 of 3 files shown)


strengejacke:
It's likely that this PR breaks tests, because CIs for glmer and glmmTMB now may change.

strengejacke:
Let's close this PR. After discussion with @mattansb, and looking at results from mixed models, I think `type = "response"` is preferable, because it averages across random effects, which is one of the strengths of the _marginaleffects_ packages.

vincentarelbundock:
Sure, that's fine, but these are separate questions.

This line of code ensures that we always default to "response" whenever there is aggregation: https://github.com/vincentarelbundock/marginaleffects/blob/main/R/sanitize_type.R#L36

So your change would not have affected marginalization at all. It would only have had an effect for `predictions()` when `by=FALSE`.

strengejacke:
I see! I hadn't realized this earlier, I though `"invlink(link)"` is always used for certain classed, independent from values of `by`.

I see three options:
1. Use `"response"`, and in return, we accept CIs outside meaningful bounds (e.g. negative in case of count response)
2. Use `"invlink(link)"`, aggregate (by `by`) and then back-transform, which gives us accurate CIs, but in return we have less accurate predictions, in particular for mixed models, where averaging across random effects doesn't work well then
3. Use a form of bias correction (not yet implemented) applied to the `"invlink(link)"` option, which - at least for the few cases I tested - comes close to `"response"`, but more accurate CIs.

My impression is when you have to decide between options 1) and 2), I would rather choose 1) right now.

vincentarelbundock:
Right, so the design choice I made in those two cases are:

1. Aggregation via `by` or `avg_*()`: Always use `response` and never do backtransformation.
2. No aggregation with `predictions()`: Backtransform when we *know* it is appropriate. For now, that is only `glm()`, but I am open to adding this for others if we are 100% sure that the backtransformation is applied only where appropriate. For example, we need an absolutely safe, 100% guaranteed way to exclude all model classes like hurdle from `glmmTMB`.

Issue #1419, opened 2025-03-22
`construct_call` and exports


vincentarelbundock:
@ngreifer,

I think your new `construct_call()` may be breaking a common use case. Do you know how we could fix the `roxygen` code to have proper access to nested function calls?

``` r
mod = lm(mpg ~ hp, mtcars)
marginaleffects::avg_slopes(mod)
#> Error in slopes(model = structure(list(coefficients = c(`(Intercept)` = 30.0988605396225, : could not find function "slopes"
```

ngreifer:
I actually just encountered this today, too. I think if you ensure the functional starts with `marginaleffects::`, e.g., `marginaleffects::slopes`, it should work okay. This is what is done in the source code for `lm()`, for example.

vincentarelbundock:
Hmmm, ok.

That will probably break my development/debugging workflow, which relies heavily on `pkgload` 

😔

vincentarelbundock:
For some reason, I can't get this to work. I tried:

1. In `avg_slopes()`, calling this: `construct_call(model, "marginaleffects::slopes")`
2. In `construct_call()`, calling `names(formals(get(calling_function, pos = asNamespace("marginaleffects")))`

PR #1422, opened 2025-03-22
Fix issue with #1419

ngreifer:
This seems to work on my end. Loading with `load_all()` to force local changes still works.


vincentarelbundock:
Thanks for looking into this, I appreciate it!

Looks like things are working well.

PR #1449, opened 2025-05-09
Improved printing when output is split

ngreifer:
I improved the printing when the output is split due to too many rows. Previously, the top and bottom wouldn't necessarily be aligned. Now they are. I also did some minor code cleaning (mostly replacing `paste(., collapse = ", ")` with the cleaner but equivalent `toString(.)`).


vincentarelbundock:
Sounds awesome. Thanks a lot!

PR #1477, opened 2025-05-30
`inferences(estimator = ...)`

vincentarelbundock:


ngreifer:
Interesting, I like it! It's not really necessary for `WeightIt` estimation functions because using `update()` on the output of, e.g., `glm_weightit()`  with a new dataset automatically re-estimates the weights with the supplied data. But as a general purpose tool I think it's a great idea.

PR #1499, opened 2025-06-21
jax experiment

vincentarelbundock:
@arcruz0 here’s a minimal integration of JAX in `predictions()`. I get about 2x speedup in one contrived example.

I think we should experiment with something like that to see if there’s real potential for massive gains. That will give us a good sense of whether this is worth the effort.

Right now, this only works for `predictions(mod)` without any argument, and only for `lm` models. Internally, we check if `getOption("jax")` is `TRUE` and if so, we use JAX to compute the Jacobian of the predictions.

If we want to make this more general, we would need to:

1.  Have a more rigorous way to tell `get_se_delta()` which specific calls are JAX-supported.
2.  Implement new functions like the one I put in `R/jax.R`.
3.  Add some safety checks to make sure the Python environment is properly configured and users get informative error messages.
4.  Maybe refactor the code to make it more like a “plugin” infrastructure.

``` r
library(microbenchmark)
library(marginaleffects)

insight::check_if_installed("reticulate")
numpy <<- reticulate::import("numpy")
jax <<- reticulate::import("jax")

N <- 100000
dat <- data.frame(
    y = rnorm(N),
    x1 = sample(letters, N, replace = TRUE),
    x2 = sample(letters, N, replace = TRUE),
    x3 = sample(letters, N, replace = TRUE)
)
mod <- lm(y ~ x1 + x2 + x3, data = dat)

p_jax <- function() {
    options(jax = TRUE)
    predictions(mod)
}
p_no_jax <- function() {
    options(jax = FALSE)
    predictions(mod)
}

a = p_jax()
b = p_no_jax()
all.equal(a$estimate, b$estimate)
#> [1] TRUE
all.equal(a$std.error, b$std.error)
#> [1] TRUE

microbenchmark(
    p_jax(),
    p_no_jax(),
    times = 25
)
#> Warning in microbenchmark(p_jax(), p_no_jax(), times = 25): less accurate
#> nanosecond times to avoid potential integer overflows
#> Unit: milliseconds
#>        expr      min       lq     mean   median       uq      max neval cld
#>     p_jax() 269.6382 276.2116 294.5163 284.4181 291.2288 350.8728    25  a 
#>  p_no_jax() 665.8196 678.2469 704.7367 692.8608 722.2983 782.5606    25   b
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -151,6 +151,7 @@ Suggests:
     rcmdcheck,
     remotes,
     reticulate,
+    reticulate,
     rmarkdown,
     rms,
     robust,
@@ -239,6 +240,7 @@ Collate:
     'inferences_fwb.R'
     'inferences_rsample.R'
     'inferences_simulation.R'
+    'jax.R'
     'matrix_apply.R'
     'mean_or_mode.R'
     'methods.R'
--- a/R/get_se_delta.R
+++ b/R/get_se_delta.R
@@ -98,6 +98,14 @@ get_se_delta <- function(
         coefs <- coefs[bnames]
     }
 
+    # get Jacobian for predictions using JAX
+    jax_flag <- isTRUE(getOption("jax", default = FALSE))
+    if (jax_flag) {
+        X <- attr(newdata, "marginaleffects_model_matrix")
+        # assigning a value to J immediately will skip the expensive computations below
+        J <- jacobian_jax_lm(X, coefs)
+    }
+
     # input: named vector of coefficients
     # output: gradient
     inner <- function(x) {
--- a/R/jax.R
+++ b/R/jax.R
@@ -1,0 +1,10 @@
+jacobian_jax_lm <- function(X, coefs) {
+    np_array <- numpy$array
+    get_predict_jax_lm <- function(coefs) {
+        X %*% coefs
+    }
+    f_jacobian <- jax$jacfwd(get_predict_jax_lm)
+    J <- f_jacobian(np_array(coefs))
+    J <- numpy$array(J)
+    return(J)
+}


arcruz0:
Thanks @vincentarelbundock, this is is really cool. I've been tinkering on [arcruz0/marginaleffects/tree/jax_experiment](https://github.com/arcruz0/marginaleffects/tree/jax_experiment). What's the best to go about this logistically? Should I submit a PR here that I can keep revising? Some comments below.

## Functionality/optimization

I added bare-bones support for average predictions via `predictions()` with the `by = ` argument. See [`get_se_delta.R#L111-120`](https://github.com/arcruz0/marginaleffects/blob/6787cb1111e30b557f5cdf7136083f9ab6bfaf69/R/get_se_delta.R#L111-L120) and [`jax.R#L58-L73`](https://github.com/arcruz0/marginaleffects/blob/6787cb1111e30b557f5cdf7136083f9ab6bfaf69/R/jax.R#L58-L75). Performance looks good (see Benchmarks below), although the pipeline obviously needs to be more robust. @vincentarelbundock I'd really appreciate thoughts on the pipeline—how would this "plugin" implementation look like, for example?

I also did more optimization, mainly via [just-in-time compilation (JIT)](https://docs.jax.dev/en/latest/jit-compilation.html) at load time. While it's harder to write JIT-compatible functions (in fact I think my implementation for average predictions is suboptimal), it's another big attraction of JAX performance-wise and seems to help a bit here. 

## Benchmarks

Below are some benchmarks ([see gist for code](https://gist.github.com/arcruz0/2e43588c968cfaa741289ccd5d3f4977)). These are linear models with $N$ observations,  $y \sim N(0,1)$, and $K$ independent variables, half of which are $x_k \sim N(0,1)$ and half of which are $x_k \sim Bernoulli(0.5)$. Average predictions are for one of the latter dummies.

The JAX backend really shines as $N$ and $K$ grow:

image

## Backend implementation

For handling the installation of Python libraries, `reticulate` has a relatively new [vignette](https://rstudio.github.io/reticulate/articles/package.html) describing a process around `reticulate::py_require()`. This is what I implemented in [`jax.R`](https://github.com/arcruz0/marginaleffects/blob/6787cb1111e30b557f5cdf7136083f9ab6bfaf69/R/jax.R) and seems to be straightforward, but it would require more testing.

As for how to call the Python libraries internally (so that the user doesn't need to do `jax <<-`, etc.), I've adopted what @cjerzak does in his JAX-powered R packages (e.g., [cjerzak/fastrerandomize](https://github.com/cjerzak/fastrerandomize-software/blob/66ab83bf4ea4fc7755df8d8ea92cbbfd8c48ce8a/fastrerandomize/R/FRR_InitializeJAX.R)), which is to create an internal environment that hosts the Python libraries and functions. 

So the function [`use_jax_backend()`](https://github.com/arcruz0/marginaleffects/blob/6787cb1111e30b557f5cdf7136083f9ab6bfaf69/R/jax.R) (1) runs `insight::check_if_installed()` for `reticulate`; (2) runs `reticulate::py_require()` to install the Python libraries if necessary; (3) loads Numpy/JAX to the internal environment; and (4) loads and JITs the internal JAX functions to said environment. The user would simply do something like this:

``` r
# remotes::install_github("arcruz0/marginaleffects@jax_experiment")
library(marginaleffects)
use_jax_backend()
#> JAX set as backend for `marginaleffects`.
#>   Run `⁠options(jax = FALSE)` to disable.
mod <- lm(mpg ~ hp + am, mtcars)
predictions(mod)
#> 
#>  Estimate Std. Error     z Pr(>|z|)     S 2.5 % 97.5 %
#>      25.4      0.818 31.05   <0.001 700.5 23.78   27.0
#>      25.4      0.818 31.05   <0.001 700.5 23.78   27.0
#>      26.4      0.850 31.06   <0.001 701.1 24.72   28.1
#>      20.1      0.775 25.93   <0.001 490.0 18.59   21.6
#>      16.3      0.677 24.03   <0.001 421.6 14.95   17.6
#> --- 22 rows omitted. See ?print.marginaleffects ---
#>      25.2      0.814 30.96   <0.001 696.8 23.61   26.8
#>      16.3      1.346 12.12   <0.001 109.9 13.68   19.0
#>      21.6      0.891 24.19   <0.001 427.0 19.81   23.3
#>      12.1      1.824  6.65   <0.001  35.0  8.56   15.7
#>      25.4      0.819 31.07   <0.001 701.5 23.84   27.0
#> Type: response
predictions(mod, by = "am")
#> 
#>  am Estimate Std. Error    z Pr(>|z|)     S 2.5 % 97.5 %
#>   0     17.1      0.667 25.7   <0.001 481.2  15.8   18.5
#>   1     24.4      0.807 30.2   <0.001 664.5  22.8   26.0
#> 
#> Type: response
```

<sup>Created on 2025-06-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

arcruz0:
That pipeline sounds great. Much cleaner than messing too much with `get_se_delta()`.

> This keeps changes to the main codebase minimal: just a conditional wrapper around a call to the user-supplied function.

Yes, this would be good. I suppose some sort of informative warning/message for when `marginaleffects_jacobian_function` is specified but not found would make sense too. Any other messages (e.g., the function received an unsupported call, etc.) can be handled by the plugin package.
Eventually there could be a `marginaleffects::use_marginaleffects_jax()` function to run `insight::check_if_installed()` on the plugin package and possibly activate it; for now I think that is not needed.

If you can implement that `marginaleffects_jacobian_function` option in `get_se_delta()`, I'm happy to move development of the JAX backend to a separate package (maybe `marginaleffectsjax`?). I'll try to be comprehensive with tests!

vincentarelbundock:
Cool. I implemented that. Closing this experimental branch and moving the conversation to https://github.com/vincentarelbundock/marginaleffects/issues/1506

PR #1515, opened 2025-07-08
Fixes to inferences()

ngreifer:
Updating for new version of `fwb`. Fixed issue where dots were not passed to `fwb()`. Added p-values to fwb, as these are now produced by `summary.fwb()` by inverting CIs. Added `conf_type` to simulation, allowing for Wald inference (including p-values); default perc is unchanged. Minor cleaning here and there to improve organization.


vincentarelbundock:
Looks fantastic! Thanks a lot!

PR #1680, opened 2026-03-05
Support nestedLogit

vincentarelbundock:


vincentarelbundock:
Thanks! Should be fixed now

Issue #1718, opened 2026-04-26
Expand "Extentions" chapter to include MCMC models


mattansb:
I've found that only the `get_predict()` function needs to be defined for this to work:

https://gist.github.com/mattansb/617e99ff73be12f7ca499fd9018e31b5

(Though we do get a warning "Unable to extract a variance-covariance matrix from this model.").

vincentarelbundock:
OK, this is amazing!

I'm totally overwhelmed with other projects now, but would be glad to include examples like thin in a vignette. Either extensions or Bayes vignette (with links in both direction).

mattansb:
Okay, I made a simple PR.

vincentarelbundock:
merged, thanks!

PR #1719, opened 2026-04-26
extending Bayes

mattansb:
Addressing #1718 

Added a sub-section in the "Extending" chapter to show-off a JAGS logistic regression (the same ideas can be used to also use `marginaleffects` for Stan etc...).

I tried to link to the "Bayes" chapter and vice versa, but I'm not sure how linking between chapters works?


vincentarelbundock:
This look amazing, thanks so much!

I really appreciate it. It's fun to be surprised by what the package can do, even after so much time spent with it...

mattansb:
From "didn't know my package can do that" to "official part of the API" ;)

PR #1725, opened 2026-05-20
[BUG] Circular import in `marginaleffects==0.5.0` breaks direct `datagrid` import

### LINKED ISSUE #1724: Circular import in `marginaleffects==0.5.0` breaks direct `datagrid` import
Hey! This is a weird bug I am facing, and this report was generated by Claude :)

I will try to create a PR

---

## Summary

In `marginaleffects==0.5.0`, importing `datagrid` as the first symbol from the package in a fresh Python process fails with an `ImportError` caused by a circular import between `marginaleffects.datagrid`, `marginaleffects.utils`, and `marginaleffects.sanitize.newdata`. In long-running sessions (e.g. Jupyter), the same cycle can leave `marginaleffects.utils` partially initialized in `sys.modules`, which surfaces later as a confusing `AttributeError: module 'marginaleffects.utils' has no attribute 'get_mode'`.

The bug is not caught by the test suite because typical test/user code imports another top-level function (such as `predictions`) before `datagrid`, which happens to load `marginaleffects.sanitize` first and primes the import graph in a working order.

## Environment

- `marginaleffects==0.5.0` (latest on PyPI as of 2026-05-20)
- Python 3.13
- macOS (also reproducible on Linux — issue is pure Python import-graph)

## Minimal reproduction

In a **fresh** Python process:

```bash
python -c "from marginaleffects import datagrid"
```

Expected: silent success.

Actual:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
    from marginaleffects import datagrid
  File ".../site-packages/marginaleffects/__init__.py", line 39, in __getattr__
    module = import_module(module_name)
  File ".../importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File ".../site-packages/marginaleffects/datagrid.py", line 6, in <module>
    import marginaleffects.utils as ut
  File ".../site-packages/marginaleffects/utils.py", line 6, in <module>
    from .sanitize.utils import validate_string_columns  # noqa: F401
  File ".../site-packages/marginaleffects/sanitize/__init__.py", line 8, in <module>
    from .newdata import sanitize_newdata
  File ".../site-packages/marginaleffects/sanitize/newdata.py", line 4, in <module>
    from ..datagrid import datagrid
ImportError: cannot import name 'datagrid' from partially initialized module
'marginaleffects.datagrid' (most likely due to a circular import)
```

Equivalent failing entry points:

```python
from marginaleffects.datagrid import datagrid
import marginaleffects.utils
import marginaleffects; marginaleffects.datagrid  # triggers __getattr__
```

## The cycle

```
marginaleffects.datagrid
        │  line 6: import marginaleffects.utils as ut
        ▼
marginaleffects.utils
        │  line 6: from .sanitize.utils import validate_string_columns
        ▼
marginaleffects.sanitize.__init__
        │  line 8: from .newdata import sanitize_newdata
        ▼
marginaleffects.sanitize.newdata
        │  line 4: from ..datagrid import datagrid   ← datagrid is still mid-load
        │  line 5: from ..utils import ingest, upcast ← utils is still mid-load
        ▼
            ImportError
```

The cycle exists at module-import time: `utils` pulls in `sanitize`, `sanitize` pulls in `newdata`, and `newdata` pulls back into `datagrid` and `utils` before either has finished executing.

## Confusing downstream symptom in long-running sessions

In notebooks where an earlier (failed) import attempt has already placed a partial `marginaleffects.utils` into `sys.modules`, the second symptom appears: `datagrid` imports "successfully" (because Python returns the cached partial module to `import marginaleffects.utils as ut`), but the cached `ut` only contains symbols defined before `utils.py`'s line 6. The later `_get_function_defaults` call then fails with:

```
AttributeError: module 'marginaleffects.utils' has no attribute 'get_mode'
```

This is much harder to attribute to a circular import than the original `ImportError`.

## Why the test suite doesn't catch this

The package `__init__.py` uses a lazy `__getattr__` for all public exports. Most user code (and presumably most tests) does something like:

```python
from marginaleffects import predictions, slopes, comparisons, datagrid
```

Resolution of `predictions` triggers `marginaleffects.predictions` → `marginaleffects.sanitize` → loads `sanitize/__init__.py` fully → which in turn imports `datagrid` and `utils` *in an order where the cycle resolves cleanly*. By the time `__getattr__("datagrid")` runs, `marginaleffects.datagrid` is already fully initialized in `sys.modules`.

The bug only manifests when `datagrid` is the **first** symbol requested from the package. A regression test that runs `python -c "from marginaleffects import datagrid"` (in a fresh subprocess for each top-level export) would catch it.

## Working workarounds (user side)

Either of these works in a fresh kernel:

```python
# Option 1: preload the sanitize package
import marginaleffects.sanitize  # noqa: F401
from marginaleffects.datagrid import datagrid

# Option 2: import any other top-level function first
from marginaleffects import predictions  # noqa: F401
from marginaleffects import datagrid
```

Both verified on `marginaleffects==0.5.0`, Python 3.13.

## Suggested upstream fix

Break the import-time cycle. Options:

1. **Lazy-import `datagrid` inside `sanitize/newdata.py`** — move `from ..datagrid import datagrid` from the module top into the `sanitize_newdata` function body. This is the smallest change and is the conventional Python remedy for this kind of cycle.
2. **Decouple `marginaleffects.utils` from `marginaleffects.sanitize`** — `utils.py` line 6 (`from .sanitize.utils import validate_string_columns`) is what kicks off the chain; if the helpers re-exported from `sanitize.utils` were instead imported lazily inside the functions that need them, or moved into a leaf module that neither package imports at module level, the cycle disappears.
3. **Add a regression test** that imports each public symbol from `_EXPORTS` in `__init__.py` from a fresh subprocess.

## Bonus: secondary issue surfaced while debugging

`marginaleffects.datagrid.datagrid` (and its helper `_detect_variable_type`) assumes the `newdata` argument is a `polars.DataFrame`; passing a `pandas.DataFrame` raises `AttributeError: 'numpy.ndarray' object has no attribute 'drop_nulls'` from inside `_detect_variable_type`, because `pandas.Series.unique()` returns a `numpy.ndarray` rather than a polars series. Either coercing with `pl.from_pandas(...)` at the top of `datagrid()` or documenting the polars-only requirement on the parameter would be a friendlier UX.

juanitorduz:
Closes https://github.com/vincentarelbundock/marginaleffects/issues/1724

Some side changes are because of the linter (I can revert them).

### DIFF
--- a/python/marginaleffects/plot/common.py
+++ b/python/marginaleffects/plot/common.py
@@ -36,7 +36,9 @@ def dt_on_condition(model, condition):
         condition_new.pop("newdata", None)
 
     if not (1 <= len(condition_new) <= 4):
-        raise ValueError(f"Length of condition must be inclusively between 1 and 4. Got: {len(condition_new)}.")
+        raise ValueError(
+            f"Length of condition must be inclusively between 1 and 4. Got: {len(condition_new)}."
+        )
 
     for key, value in to_datagrid.items():
         variable_type = model.get_variable_type(key)
@@ -53,7 +55,9 @@ def dt_on_condition(model, condition):
                 else modeldata[key].unique().sort().to_list()
             )
             if len(to_datagrid[key]) > 10:
-                raise ValueError(f"Character type variables of more than 10 unique values are not supported. {key} variable has {len(to_datagrid[key])} unique values.")
+                raise ValueError(
+                    f"Character type variables of more than 10 unique values are not supported. {key} variable has {len(to_datagrid[key])} unique values."
+                )
 
         elif variable_type in ["boolean", "binary"]:
             if to_datagrid[key] is None:
@@ -127,7 +131,9 @@ def validate_plot_args(condition, by, newdata, wts):
     if wts is not None and not by:
         raise ValueError("The `wts` argument requires a `by` argument.")
     if not ((condition is None and by) or (condition is not None and not by)):
-        raise ValueError("One of the `condition` and `by` arguments must be supplied, but not both.")
+        raise ValueError(
+            "One of the `condition` and `by` arguments must be supplied, but not both."
+        )
 
 
 def extract_var_list(condition, by):
@@ -147,7 +153,9 @@ def extract_var_list(condition, by):
     var_list = [x for x in var_list if x not in ["newdata", "model"]]
 
     if len(var_list) >= 5:
-        raise ValueError("The `condition` and `by` arguments can have a max length of 4.")
+        raise ValueError(
+            "The `condition` and `by` arguments can have a max length of 4."
+        )
 
     return var_list
 
--- a/python/marginaleffects/sanitize/newdata.py
+++ b/python/marginaleffects/sanitize/newdata.py
@@ -1,12 +1,16 @@
 import numpy as np
 import polars as pl
 
-from ..datagrid import datagrid
-from ..utils import ingest, upcast
 from ..formula import listwise_deletion
 
 
 def sanitize_newdata(model, newdata, wts, by=[]):
+    # Lazy imports to break the `datagrid -> utils -> sanitize -> newdata -> ...`
+    # circular import that fires when `datagrid` is the first symbol pulled from
+    # marginaleffects in a fresh interpreter (see GH #1724).
+    from ..datagrid import datagrid
+    from ..utils import ingest, upcast
+
     modeldata = model.get_modeldata()
 
     if newdata is None:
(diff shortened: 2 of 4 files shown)


vincentarelbundock:
Thanks a lot for the fix @juanitorduz ! I really appreciate you taking the time!

Version 0.5.1 should be available on PyPI now.

juanitorduz:
> Thanks a lot for the fix @juanitorduz ! I really appreciate you taking the time!
> 
> Version 0.5.1 should be available on PyPI now.

Of course! I am giving a presentation about `marginaleffects` XD

Thanks for cutting a release :)

vincentarelbundock:
Oh, that's exciting!

PR #1729, opened 2026-06-15
Support statsmodels Nominal/Ordinal GEE in predictions/comparisons/slopes

huntermills707:
NominalGEE and OrdinalGEE store coefficients in an expanded multi-equation form whose inherited GLM predict cannot consume an ordinary (n, k) design, so predictions/comparisons/slopes previously failed with a cryptic numpy shape error. Reconstruct category probabilities from the plain design:

- NominalGEE: multinomial-logit inverse link over per-category linear predictors.
- OrdinalGEE: cumulative-logit survival probabilities P(Y > c_j) turned into category probabilities by successive differences. 

Both feed the existing 2-D rowid x group path used for MNLogit. Dispatch is a name->reconstructor registry (matching the family_map/link_map style in this module) rather than a class-name if/elif chain, so new models are one map entry plus one method.

This change does meaningfully enlarge `python/marginaleffects/statsmodels/model.py`.

Open to suggestions (maybe punt the predict functions into another module and out of the ModelStatsmodels class). 
Also open to rejection if too niche.

### DIFF
--- a/python/marginaleffects/statsmodels/model.py
+++ b/python/marginaleffects/statsmodels/model.py
@@ -10,6 +10,15 @@ from ..sanitize.utils import validate_types
 
 
 class ModelStatsmodels(ModelAbstract):
+    # A few statsmodels models store coefficients in an expanded multi-equation
+    # form whose own `predict` cannot consume an ordinary (n, k) design matrix.
+    # Map the model class name to a method that reconstructs predictions from
+    # the plain design; every other model delegates to its native `predict`.
+    _PREDICT_RECONSTRUCTORS = {
+        "NominalGEE": "_predict_nominal_gee",
+        "OrdinalGEE": "_predict_ordinal_gee",
+    }
+
     def __init__(self, model, vault=None):
         if vault is None:
             vault = ModelVault()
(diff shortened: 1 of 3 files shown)


vincentarelbundock:
Thanks a lot for this @huntermills707 , I really appreciate it!

It's kind of a crazy week for me, so I'm not sure exactly when I'll have time to try it out, but I'll do it as soon as possible.

huntermills707:
No rush. 
I have a personal life goal to never be in a hurry.
Can't say I am very successful at it, but I try.
Only look at this when it is convenient.

huntermills707:
No that I have stood back and looked at this more critically, the fundamental issue is that statsmodels dropped the predict methods for Ordinal and Nominal GEE a while back during a refactor.

This is a fix that makes more sense to land in statsmodels than it does here.

If it does not land there, it probably still does not make sense to land here.

Probably going to close this, and attempt to make this change there.

vincentarelbundock:
Thanks for the follow-up!

Yes, I agree that this makes more sense upstream. I don't think we want to be in the business of defining predict methods for all the models we could potentailly support. That's a very deep rabbit hole. Will close this but follow upstream developments with interest.

Thanks!

PR #1746, opened 2026-08-16
Analytic Jacobians for slopes and nonlinear contrasts

vincentarelbundock:
## Summary

The exact-derivative path covered only `difference` comparisons. That left `slopes()` and `avg_slopes()` — the package's namesake — falling back to finite differences over the coefficient vector on **every** model, along with every ratio-family contrast.

This extends the closed-form Jacobian in both R and Python to `dydx`, `ratio`, `lnratio`, `lnor`, `lift`, and `expdydx`, plus their `avg` and weighted variants.

Coverage before/after on a handful of representative calls (`analytic=TRUE` means the closed form was used):

| call | before | after |
|---|---|---|
| `slopes(lm)` / `avg_slopes(lm)` | ✗ | ✓ |
| `slopes(glm, type="response")` | ✗ | ✓ |
| `avg_slopes(glm, by=...)` | ✗ | ✓ |
| `comparisons(..., comparison="ratio"/"lnratio"/"lift")` | ✗ | ✓ |
| `avg_comparisons(glm, comparison="lnor")` | ✗ | ✓ |
| `avg_slopes(slope="eyex")` | ✗ | ✗ (by design) |

## Approach

Row-level gradients are the chain rule applied to the corresponding entries in `comparison_function_dict` (R) / `estimands` (Python). The `avg` forms split into two families:

- keys whose average is the average of row-level comparisons (`difference`, `dydx`, `expdydx`) → weighted mean of the row-level gradients;
- keys that aggregate the predictions *before* applying a nonlinear function (`ratioavg`, `lnratioavg`, `lnoravg`, `liftavg`) → gradient evaluated at the aggregated predictions.

`lift` shares the `ratio` gradient, since `lift = hi/lo - 1`.

**The x-derivative stays numerical.** Only the coefficient-side gradient is closed-form, so the two-point `(hi, lo)` structure is untouched and formula terms like `I()`, `poly()`, and splines need no special handling — the analytic path only ever sees an already-built design matrix. Elasticities (`eyex`, `eydx`, `dyex`) keep the fallback because they depend on the fitted response and need a product rule.

## Python: response-scale GLMs

Python gains the chain rule R already had. The inverse-link derivative is read off the fitted statsmodels link object via `inverse_deriv()` rather than switching on a link-name string, so custom and unlisted links work without an enumeration. Adapters opt in with a new `get_link_functions()` returning `(linkinv, mu_eta)`.

## Accuracy

The replay guard is unchanged: a closed-form Jacobian is returned only if replaying the plan reproduces the point estimates. Both packages now additionally reject a non-finite Jacobian and fall back rather than propagating it into the variance.

Standard errors for the newly covered estimands can move in about the 5th significant digit. **The analytic values are the accurate ones** — the tests pin them to closed forms computed outside the package:

- logit AME: analytic matches the exact form to `2e-10`; finite differences were off by `2e-5`
- unit-level Poisson slopes: analytic to `4e-9`; finite differences off by `3.3e-3`

A pleasant check that fell out: `lnor` on a logit response scale must collapse exactly to `X_hi - X_lo`, since log-odds are linear in the coefficients. That is asserted directly, with no numerical reference.

## Test changes (please review)

Two existing tests encoded the narrower whitelist and had to change:

- `test-jacobian-analytic.R` — the cumulative finite-difference call counters shift by one, because the `comparison = "ratio"` call no longer reaches the fallback. Every subsequent expected count is decremented; the rest of the test (offsets, rank deficiency, custom functions, the global opt-out) is unchanged in meaning.
- `test-by.R` — compares a custom `comparison` function (still finite differences) against `"ratioavg"` (now exact). They agree to `5.3e-8`, just above tinytest's default `sqrt(.Machine$double.eps)`, so that one assertion gets an explicit tolerance.
- `test_autodiff.py` — the "forcing autodiff warns when unsupported" test used `comparison="lift"`, which the analytic path now handles before autodiff is reached. Switched to `eyex`, which is genuinely outside both paths.

## Testing

- R: full suite `All ok, 2559 results`
- Python: `350 passed` (6 pre-existing plot image-regression failures deselected; they fail identically on `main`)
- `ruff` error count unchanged from the 238 pre-existing baseline

New tests: 45 R assertions in `test-jacobian-analytic.R` and 6 Python tests in `test_inference_analytic.py`, covering every new key, weighted averages, both aggregation branches, GLM response scale, and the fallback cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/NEWS.md
+++ b/python/NEWS.md
@@ -1,3 +1,21 @@
+# 0.6.2
+
+Performance:
+
+* Analytic Jacobians now cover GLMs on the response scale. The inverse-link
+  derivative is read from the fitted statsmodels link object, so logit, probit,
+  Poisson, and any other link exposing `inverse_deriv()` are supported without
+  an explicit list.
+* `slopes()` and `avg_slopes()` now use exact analytic Jacobians, along with the
+  `ratio`, `lnratio`, `lnor`, `lift`, and `expdydx` comparisons and their `avg`
+  and weighted variants. Elasticities (`eyex`, `eydx`, `dyex`) depend on the
+  fitted response and retain the numerical fallback.
+* Standard errors for these estimands can differ from previous versions in about
+  the 5th significant digit, because the closed form replaces a finite-difference
+  approximation. The analytic values are the accurate ones.
+* Model adapters may implement `get_link_functions()` to return
+  `(linkinv, mu_eta)` and opt into the response-scale analytic path.
+
 # 0.6.1
 
 Breaking changes:
--- a/python/marginaleffects/classes/model.py
+++ b/python/marginaleffects/classes/model.py
@@ -144,6 +144,13 @@ class ModelAbstract(ModelValidation, ABC):
     def get_autodiff_args(self):
         return None
 
+    # Adapters predicting on a response scale return `(linkinv, mu_eta)` here.
+    # `mu_eta` is the derivative of the inverse link, which turns a link-scale
+    # design matrix into a response-scale Jacobian. `None` keeps the analytic
+    # path from claiming the model.
+    def get_link_functions(self):
+        return None
+
     def get_df(self) -> float:
         return np.inf
 
--- a/python/marginaleffects/inference/analytic.py
+++ b/python/marginaleffects/inference/analytic.py
@@ -1,4 +1,4 @@
-"""Analytic Jacobians for model-matrix based linear predictions."""
+"""Analytic Jacobians for model-matrix based predictions."""
 
 from dataclasses import dataclass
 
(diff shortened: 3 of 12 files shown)


vincentarelbundock:
Closing: superseded. The R analytic-Jacobian path has since been rebuilt around stage-by-stage composition (`jacobian_stage.R`, the `(pre, post)` seam in the replay plans, and the syntax-tree hypothesis compiler), and it now covers every comparison key this PR added plus `dyex`.

The Python side is not obsolete but should follow that architecture rather than this one — porting the wider key table into the current all-or-nothing structure would grow the surface that later has to be unwound to introduce the stage seam. That work is specced out in `SPEC-python-jacobian-parity.md`, which carries over the `get_link_functions()` adapter hook from here (reading `inverse_deriv` off the statsmodels link object instead of enumerating link names) as work item W2.

PR #1749, opened 2026-08-17
Fall back to the model matrix when predict() fails (#1748)

### LINKED ISSUE #1748: Rank-deficient model with `vcov=FALSE` and `qr$qr` removed
Passing a rank-deficient `lm` or `glm` model to `avg_predictions` with `vcov=FALSE` errors when the fit object has the `qr$qr` element removed. This does not error if `vcov=FALSE` or if the `qr$qr` element is kept in the fit object.

```
library(marginaleffects)
df <- mtcars

# create rank deficiency
df$hp2 <- df$hp + 2
fit <- lm(mpg ~ hp+hp2, data = df)

# set qr element to NULL to preserve memory
fit$qr$qr <- NULL

avg_predictions(model=fit,
                vcov=FALSE
)

```

```
Error: Unable to compute predicted values with this model. This error can arise when `get_modeldata()` is unable to extract the dataset from the model object, or when the original data frame (usually in the global environment) was modified since fitting the model. Make sure the original data object is unchanged and/or try to supply a different dataset to the `newdata` argument.

In addition, this error message was raised:
'from' must be a finite number

Bug Tracker: https://github.com/vincentarelbundock/marginaleffects/issues

```

`sessionInfo()`

```
R version 4.4.2 (2024-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows Server 2019 x64 (build 17763)

Matrix products: default


locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

other attached packages:
[1] marginaleffects_0.32.0.9995

loaded via a namespace (and not attached):
 [1] compiler_4.4.2    backports_1.5.1   cli_3.6.6         generics_0.1.4    tools_4.4.2       rstudioapi_0.19.0 yaml_2.3.12      
 [8] remotes_2.5.0     data.table_1.18.4 insight_1.5.2     checkmate_2.3.4   rlang_1.3.0       renv_1.1.5     
```

vincentarelbundock:
Closes #1748.

## Problem

Deleting `qr$qr` from an `lm`/`glm` fit to save memory breaks `stats::predict()` itself:

```r
fit <- lm(mpg ~ hp + hp2, data = df)
fit$qr$qr <- NULL
predict(fit, newdata = df)
#> Error: 'from' must be a finite number   (via min(dim(R)) -> Inf)
```

marginaleffects used to sidestep this: `get_predict.lm()` computes `X %*% beta` from the cached model matrix and never touches the QR decomposition. #1741 made that cache conditional on delta-method inference (`r/R/predictions.R:260`, `r/R/get_comparisons_data.R:175`), so `vcov = FALSE` calls now fall through to the failing `predict()` method and surface as:

> Unable to compute predicted values with this model. […] 'from' must be a finite number

This affects `predictions()`, `comparisons()`, and `slopes()`. Rank deficiency is incidental — any stripped fit fails the same way.

## Fix

Rather than restoring the unconditional cache and giving up #1741's performance win, make the matrix path a real fallback instead of an incidental one. When `stats::predict()` raises an error, `get_predict.lm()` / `get_predict.glm()` build the model matrix on demand and retry through `X %*% beta`.

- The matrix is still not built on the `vcov = FALSE` happy path — construction only happens after `predict()` has already failed, which never happens for a well-formed fit.
- If no usable matrix can be built, the **original** `predict()` error is re-raised, so unrelated failures (stale modeldata, factor-level mismatches) keep their own message.
- `add_model_matrix_attribute()`'s existing guards still apply, so the retry can't silently produce wrong numbers for models where `X %*% beta` isn't the right computation — it returns no attribute and we re-raise.
- The glm retry is restricted to the link and response scales.

`comparison_plan_build()` reached `get_predict.lm()` with `mfx = NULL`, so `mfx` is now threaded through `predictions_hi_lo()` and the `pred_or` call; `predictions()` already passed it.

## Tests

New cases in `test-bugfix.R` cover `lm` and `glm` (response and link) across predictions/comparisons/slopes with `qr$qr` stripped, asserting the estimates match the un-stripped fit exactly, plus one asserting unrelated `predict()` failures still report their own cause.

`make r-testseq`: All ok, 2521 results.

### DIFF
--- a/r/DESCRIPTION
+++ b/r/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: marginaleffects
 Title: Predictions, Comparisons, Slopes, Marginal Means, and Hypothesis Tests
-Version: 0.32.0.9995
+Version: 0.32.0.9996
 Authors@R:
     c(person(given = "Vincent",
              family = "Arel-Bundock",
--- a/r/NEWS.md
+++ b/r/NEWS.md
@@ -35,6 +35,7 @@ Performance:
 
 Bug fixes:
 
+* `lm` and `glm` estimates no longer fail when the fit object was stripped of its `qr$qr` element to save memory. Linear predictions are now computed from the model matrix when `stats::predict()` fails. Thanks to @trose64 for report #1748.
 * `by` aggregation with posterior or bootstrap draws no longer flattens the draws matrix when a `by` data frame omits some term/group combinations.
 * `predictions()` now assigns the correct `rowid`, `type`, and `estimate` column names when model predictions are returned as a bare vector.
 * `equivalence` tests now support vector-valued degrees of freedom.
--- a/r/R/comparisons_plan.R
+++ b/r/R/comparisons_plan.R
@@ -153,7 +153,7 @@ comparison_plan_build <- function(
     newdata <- mfx@newdata
     model <- if (is.null(model_perturbed)) mfx@model else model_perturbed
 
-    predictions <- predictions_hi_lo(model, lo, hi, type, ...)
+    predictions <- predictions_hi_lo(model, lo, hi, type, mfx = mfx, ...)
     pred_lo <- predictions$pred_lo
     pred_hi <- predictions$pred_hi
     eta_lo <- attr(pred_lo, "marginaleffects_linear_predictor")
@@ -181,6 +181,7 @@ comparison_plan_build <- function(
             model,
             type = type,
             newdata = original,
+            mfx = mfx,
             ...
         )
     } else {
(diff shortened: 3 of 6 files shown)


trose64:
@vincentarelbundock - thanks for the super fast fix!

vincentarelbundock:
superseded by https://github.com/vincentarelbundock/marginaleffects/pull/1751

Issue #1735, opened 2026-07-03
`hypothesis` formulas could be converted to matrix when appropriate to go forward in lowering rules


vincentarelbundock:
allows autodiff on `hypothesis`

PR #1751, opened 2026-08-19
Compose Jacobians stage by stage instead of all at once

### LINKED ISSUE #1750: avg_comparison vs manual g-computation can produce different standard errors
My understanding is that avg_comparisons uses g-computation under the hood, so manual g-compuation using avg_predictions should give the same results. I am trying to use manual g-compuation so I can estimate general constrasts (e.g. avg(A+B) vs C), not only pairwise comparisons.

I noticed that when I used manual g-computation, and increased the sample size, the standard errors become different. Relative error can be 50%. It might be a rounding error propagating not effecting p-values, but perhaps good to investigate. 

If there was a capability to use general contrasts in avg_comparisons, that would be nice.

```
### Create data set
# Simulate data for varying sized units
set.seed(12L)
n_units = 10^6
data = data.frame(
  id = factor(1:n_units),
  size = sample(x = 10^3:10^7, size = n_units), # size of the unit
  p = rbeta(n = n_units, shape1 = 40, shape2 = 40)) # conversion / share

# Assign 4 treatments randomly
k = 4
data = data %>%
  dplyr::mutate(trt = sample(rep(c("ctrl","testA", "testB", "testC"), n_units/k)))

# Simulate n (total) and y (success) 
# lift p for test group by eff_size
# n is lognormal and depends on size
# y is binomial with probability p and size n
eff_size = 0.05
data = data %>%
  dplyr::mutate(
    p_lift = p * ifelse(trt == "testA", 1+eff_size, 
                        ifelse(trt == "testB", 1+2*eff_size,
                               ifelse(trt == "testC", 1+3*eff_size, 1))),
    n = round(rlnorm(n = n(), meanlog = -10, sdlog = 0.4) * size) + 1,
    y = rbinom(n = n(), size = n, prob = p_lift),
    x = 0.1+2*log(p)+rnorm(n(), 0, 1)) # covariate

# Level treatment
data = data %>% mutate(trt = relevel(as.factor(trt), ref = "ctrl"))

### Fit GLM model
model = glm(data = data, formula = "cbind(y, n-y) ~ trt", family = "binomial")

### Estimate contrasts
# Trt levels and contrasts
var = "trt"
var_levels = c("ctrl", "testA", "testB", "testC")
contrast_mx = matrix(
  c(-1, -1,  0, -1,  0,  0,
    1,  0, -1,  0, -1,  0,
    0,  1,  1,  0,  0, -1,
    0,  0,  0,  1,  1,  1), nrow = 4, byrow = TRUE)
rownames(contrast_mx) = var_levels
colnames(contrast_mx) = c("testA - ctrl", "testB - ctrl", "testB - testA", 
                          "testC - ctrl", "testC - testA", "testC - testB")


# avg_comparison all pairwise
eff_comp = marginaleffects::avg_comparisons(
  model = model,
  variables = setNames(list("pairwise"), var),
  type = "response",
  conf_level = 0.95,
  vcov = TRUE)

# gcomp manual
eff_gcomp = marginaleffects::avg_predictions(
  model = model, 
  type = "response",
  variables = setNames(list(var_levels), var),
  by = var,
  hypothesis = contrast_mx,
  conf_level = 0.95,
  vcov = TRUE)

### Differences
# Error in estimate
abs(eff_comp$estimate-eff_gcomp$estimate)
[1] 9.143797e-13 6.226797e-12 7.141177e-12 1.696121e-11 1.787559e-11 1.073441e-11

# Error in std.error
abs(eff_comp$std.error-eff_gcomp$std.error)
[1] 1.486783e-05 3.389385e-05 1.982944e-05 4.431031e-05 2.697474e-05 1.456325e-06

# Rel. error in estimate
abs(eff_comp$estimate-eff_gcomp$estimate)/eff_comp$estimate
[1] 3.659440e-11 1.237653e-10 2.819875e-10 2.256530e-10 3.562426e-10 4.319041e-10

# Rel. error in std.error
abs(eff_comp$std.error-eff_gcomp$std.error)/eff_comp$std.error
[1] 0.16524736 0.37728381 0.22076268 0.49515570 0.30148352 0.01630169
```

vincentarelbundock:
Fixes #1750. Implements #1735.

## The bug

`avg_predictions()` with a contrast matrix and `avg_comparisons()` are the same estimand, but their standard errors diverged as the sample grew — by up to 50% at a million rows in the reporter's example, while the point estimates still agreed to 1e-11.

It is not a statistical difference. A reduced reprex isolates it to the row count alone, holding total Fisher information fixed:

```r
mk <- function(N, total = 2e7) {
  n <- total / (2 * N)
  data.frame(trt = factor(rep(c("ctrl","test"), each = N)), n = n,
             y = c(rep(round(n * 0.50), N), rep(round(n * 0.55), N)))
}
d <- mk(N); m <- glm(cbind(y, n - y) ~ trt, data = d, family = binomial)
avg_comparisons(m, variables = "trt")$std.error
avg_predictions(m, by = "trt", hypothesis = "b2 - b1 = 0")$std.error
```

| N | comparisons | gcomp | rel. diff |
|---|---|---|---|
| 1 | 2.2305e-4 | 2.2305e-4 | 0 |
| 1e3 | 2.2305e-4 | 2.2313e-4 | 3.9e-4 |
| 1e5 | 2.2305e-4 | 2.2254e-4 | 2.3e-3 |

Same data, same model, same information — the error is driven purely by how many numbers get averaged.

## Cause

The delta method differentiated the whole composition at once: perturb a coefficient, re-run `predict()` over every row, re-average, apply the hypothesis. Two consequences.

`avg_comparisons()` differences predictions row by row and then averages, so roundoff cancels inside each row. `avg_predictions()` + hypothesis averages ~0.5-magnitude predictions over N rows *first* and only then differences, so accumulated summation roundoff of order `N * .Machine$double.eps` survives into the numerator — and is then divided by a step of `|beta| * sqrt(.Machine$double.eps)`, about 1e-9. Then the hypothesis differences two nearly-equal group means, shrinking the signal while the noise stays put.

The second consequence is structural: `get_se_delta.R` discarded an already-computed Jacobian whenever a hypothesis was present, and `get_jacobian_analytic()` bailed entirely unless the hypothesis was a literal matrix.

## Fix: compose the pipeline stage by stage

Every estimand is a pipeline — `beta -> eta -> predictions -> comparison -> aggregation -> hypothesis` — and the Jacobian of the pipeline is the product of the stage Jacobians. Only the first stage touches the model and the data; the rest is cheap arithmetic on numbers already computed. The analytic Jacobian file already composed two of these links by hand; the eligibility check was a single conjunction, so one unsupported stage discarded all of them.

`R/jacobian_stage.R` adds the missing machinery, and eligibility becomes per-stage:

- **Hypothesis.** A linear hypothesis is recovered as an exact contrast matrix by probing the closure at basis vectors and then *proving* linearity against full-size directions (#1735). This covers string hypotheses and the formula shortcuts that `hypothesis_compile_formula_matrix()` cannot build, and unlocks the autodiff lowering rules, which need a literal matrix rather than a closure. Affine maps like `"b1 - 0.5"` are deliberately not promoted — they have a derivative but cannot be written as `crossprod(H, est)`. Anything else is differentiated through its own closure and composed by the chain rule.

- **Comparison.** Only `difference`, `differenceavg`, and `differenceavgwts` had closed-form derivatives. The comparison stage is now probed directly: row-wise groups have a diagonal Jacobian, so one pair of evaluations reads off the whole diagonal, and aggregated groups whose value depends on the inputs only through a mean have a constant gradient. Neither structure is assumed — both are verified by perturbing pseudo-random subsets and checking the response against what the assumed structure predicts, which fails immediately under off-diagonal coupling or a non-uniform gradient.

- **`need_y`.** This flag is set for every custom comparison function, because such a function *might* accept the observed outcome. Whether any actually does is recorded per group, and that is the condition that matters for a derivative.

- **Numeric fallback.** When no exact Jacobian is available, differentiation now stops before the hypothesis and composes it afterwards, so the finite difference stays on a well-scaled quantity instead of an average of many predictions. Fails closed: an unusable composition redoes the original whole-pipeline derivative rather than shipping a partial one.

Nothing is approximated silently. Every probe that relies on a structural assumption returns `NULL` when verification fails, and the caller falls back to exactly what it did before. `options(marginaleffects_hypothesis_promote = FALSE)` disables promotion; `marginaleffects_stage_probe_max_dim` bounds probe width so a very wide stage cannot become slower than the fallback it replaces.

## Results

Accuracy, against a hand-written `sqrt(g' V g)`:

| N | before | after |
|---|---|---|
| 1e3 | 5.5e-07 | 3.9e-15 |
| 1e5 | 2.2e-05 | 6.7e-13 |
| 1e6 | 1.4e-04 | 6.1e-12 |

Speed, `avg_comparisons()` at 121 coefficients and 1e5 rows:

| comparison | before | after |
|---|---|---|
| `difference` | 0.33 s | 0.17 s |
| `ratio` | 2.14 s | 0.40 s |
| `lnratio` | 1.95 s | 0.44 s |
| `lnor` | 2.29 s | 0.45 s |
| `ratioavg` | 2.05 s | 0.44 s |

The gap widens with coefficient count, since finite differences cost O(k) prediction passes and the analytic path costs O(1).

## Testing

New `inst/tinytest/test-jacobian-stage.R`: 93 assertions in a plain run, 132 with autodiff enabled. Two design points drove it.

**Setting an option is not evidence that a path ran.** `get_jacobian_analytic()` returns `NULL` on any unsupported stage and `autodiff_try()` returns `NULL` when lowering fails; both fall silently through to finite differences. A suite that only set options could be entirely green while testing the same fallback three times — which is exactly what a regression narrowing the analytic path would produce. A new `jacobian_method` slot records the path that actually produced the Jacobian (`components(x, "jacobian_method")` returns `"analytic"`, `"autodiff"`, `"numeric"`, or `"custom"`), and every test asserts it.

**The three paths are not oracles of equal quality.** Finite differences are the least accurate and are the method under suspicion, so they are screened loosely rather than used as a reference; analytic and autodiff are compared tightly against each other. Because all three share `std_error_from_jacobian()` and the same aggregation weights, cross-path agreement has a blind spot, so four independent oracles are written out by hand in the test file: a two-arm risk difference, a linear-model contrast whose Jacobian is a difference of model matrix rows, a hand-differentiated ratio of averaged predictions, and a nonlinear hypothesis. The sample-size axis is tested explicitly, since #1750 is invisible at N = 100.

## Note on an existing test

`test-jacobian-analytic.R` counts finite-difference fallbacks cumulatively. Two of the estimands it counted — `comparison = "ratio"` and a custom `comparison` function — no longer fall back, so the expected counts shift down by two from that point on. Only the numbers changed; the test's intent is intact, and offsets, rank deficiency, unsupported subclasses, and the explicit opt-out all still fall back exactly as before. Both newly-analytic cases were checked against independent oracles first:

```
oracle   est 0.985272591566  se 0.003910212920
package  est 0.985272591566  se 0.003910212921
```

`test-prediction-plan.R` asserted that `~ I(f(x)) | segment` with `f(x) = c(first = x[1], total = sum(x))` stays a closure. That map is linear, so it is now promoted; the recovered matrix reproduces the closure exactly:

```
     [,1] [,2] [,3] [,4]
[1,]    1    1    0    0
[2,]    0    1    0    0
[3,]    0    0    1    1
[4,]    0    0    0    1
apply      : 2 6 8 24
crossprod H: 2 6 8 24
```

The nonlinear `ratio ~ sequential | segment` in the same file is still not promoted, and that assertion is untouched.

## Autodiff

Run with `MARGINALEFFECTS_AUTODIFF=1`, the new file is 132 assertions. 11 of the 14 grid cases lower to JAX and are compared against the analytic path at `1e-9`; the list is pinned in the test as a regression guard, since a case that stops lowering falls back silently and still returns a plausible number.

#1735 is asserted at the mechanism level rather than by result:

```
promote=TRUE  -> autodiff se 0.0288358974
promote=FALSE -> numeric  se 0.0288359009
                 warning: "Automatic differentiation with JAX does not
                 support this form of the `hypothesis` argument."
```


vincentarelbundock:
Pushed c1e8020d, which responds to both adversarial reviews of this branch's uncertainty machinery.

**Critical fix**: the comparison-stage gradient probe used one shared step scaled to the largest prediction, corrupting ratio/lift/lnratio/lnor standard errors by up to ~180x on predictions spanning 4+ orders of magnitude, silently labeled `analytic`. Reproduced at 64x on a 14-nat Poisson span. Built-in comparisons now use hand-derived closed-form stage derivatives (verified against Richardson to 3.9e-7 and exact oracles to 4.3e-9); custom closures fall back to the numeric path, since probing arbitrary code cannot prove the structure the composition relies on. Hypothesis promotion now requires a syntactic linearity proof from the parsed string; function/formula closures never promote.

Everything else in both reviews is addressed: custom-Jacobian honoring + provenance, vcov name alignment at intake, autodiff(TRUE) priority (both autodiff suites pass again, Make targets now fail on failures), simulation Wald null/direction, Python `by=True` weights, zero-SE semantics, `stata`→HC1, rq/ols offset detection, NA-plan rejection, element-wise replay guards, Inf rejection, joint-test name guard, draws-dedup alignment, `side` typo. One review claim refuted: `fdforward` is the package-wide numderiv default, not specific to `hypotheses()`.

Verification: 2467 R assertions pass (full suite minus plots), 111 autodiff assertions pass with JAX, 334 Python tests pass, plus ~50 new regression tests in test-bugfix.R covering each finding. An independent model review of the diff verified all 30 closed-form gradients numerically (≤2.3e-10) and found no blocking issues; its hardening suggestions (join-order safety in by.py, classifier top-level `=` restriction, finite constant coefficients, provenance through composed Jacobians) are included.

vincentarelbundock:
Pushed 3c6249b3, implementing the reviews' revised performance recommendations: linear hypothesis matrices are now compiled directly from the parse tree (basis-vector recovery deleted — the walk that proves linearity also produces H, with a single matvec self-check), the covariance matrix is no longer copied when already in Jacobian order, custom-closure comparison plans are rejected before any model-matrix work, and joint tests validate covariance/parameter dimensions. All targeted suites pass (15 R test files, 3 JAX-enabled files, exact-integer H matrices verified against pinned values); a full-suite sweep is running and I'll follow up if it surfaces anything.

vincentarelbundock:
Attribution note for reviewers: c47b026c ("Replay guard: matching NA patterns and factor predictions are agreement") contains more than its message says. Besides the guard fix, it carries the plan_replay.R slice of the in-progress autodiff/JAX removal (deletion of the autodiff_try dispatch block), which belongs to work another session had uncommitted in the shared working tree when I staged the whole file. The code is correct; only the commit boundary is wrong. The rest of the autodiff removal will arrive in its own commit. If you prefer clean history, rebasing that commit apart is your call — both sessions are aware.

PR #1755, opened 2026-08-21
Refine IPW explanation and code in ipw.qmd

ngreifer:
Clarified the description of the IPW technique and improved the flow of the text. Removed `wts` from `avg_comparisons()` call.


vincentarelbundock:
Thanks!

PR #1756, opened 2026-08-21
Modify matching description and update function usage

ngreifer:
Updated description of matching method and corrected function calls, in particular removing `wts` from `avg_comparisons()`.


vincentarelbundock:
Thanks, I appreciate it!

PR #1766, opened 2026-09-10
Fix bugs found by jarl's unused_object rule

vincentarelbundock:
Follow-up to #1764, where @etiennebacher noted that `jarl check r/R -s unused_object` reports ~30 unused objects in `r/R` and suggested investigating. This fixes the ones that turned out to be real bugs.

## Real bugs

**`get_vcov.systemfit()` and `get_vcov.stpm2()` referenced an undeclared `vcov`.** Neither method had `vcov` in its signature, so R resolved the symbol lexically up to `package:stats`:

```
resolves to: stats
identical to stats::vcov: TRUE
systemfit formals: model, ...
```

`sanitize_vcov(model, stats::vcov)` then took its `check_function()` branch and actually *called* `stats::vcov(model)`, discarding the result — a wasted covariance computation on every call, and any error there surfaced as `sanitize_vcov()`'s "The function supplied to the `vcov` argument raised an error" message. The user's `vcov=` request was also silently swallowed by `...`. Both methods (plus the `pstpm2`, `gsm`, `aft` aliases) now declare `vcov = NULL`.

**`sanitize_newdata()` discarded the mlogit conversion.** The `mlogit` branch assigned `as.data.table(newdata)` to `out`, but the function returns `newdata`, so the conversion the comment above it describes never happened.

## Cleanup

Eight other `get_vcov()` methods assigned `sanitize_vcov()`'s result to an unread `vcov` variable. They hardcode their own covariance extraction and only want the validation side effect, so the call is now made directly. The assignment also obscured the fact that a user-supplied `vcov` *function* gets called and thrown away.

Two `download.file()` calls got explicit `mode = "wb"`: `modelarchive_model()` downloads `.rds` files, which corrupt in text mode on Windows.

## Not changed

The remaining `unused_object` reports are false positives or harmless:

- data.table `..idx_lo` / `..out_cols` / `..cols` column selection (`get_comparisons_data.R`, `comparisons_plan.R`, `matrix_apply.R`) — invisible to static analysis
- `comparison_gradient_excluded` is used by `test-comparison-registry.R` via `:::`
- `sanitize_vcov.R:94`'s `out <- NULL` is a defensive init before a `tryCatch`
- `slopes.R:251` re-sanitizes `vcov`, but `slopes()` delegates to `comparisons()`, which does the same thing at line 234
- the rest are dead leftovers with no behavioural effect (`sanitize_variables.R`, `datagrid.R`, `hypothesis_formula.R`, `methods_gamlss.R`, …)

## Testing

`test-pkg-mlogit.R` (12), `test-pkg-survfit.R` (13, covers systemfit), `test-pkg-rstpm2.R` (2), `test-vcov.R` (25), `test-predictions.R` (59), `test-comparisons.R` (17), `test-slopes.R` (4), and `test-hypotheses.R` (71) all pass. The full sequential run is not possible in this environment — it halts on missing Suggests (`margins`, `aod`, `dbarts`, `parsnip`, `rms`, `afex`, `mhurdle`, `MCMCglmm`, `brms`), so the `get_vcov()` methods for those classes were not exercised locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01J4K95z19Z61Btxg4rgmr5o


vincentarelbundock:
Thanks a lot for this @etiennebacher

